PR #14, opened 2020-04-21
added SVG file for purple logo

jacobwilliams:
If you are interested, I added an Inkscape SVG file of the purple version of the logo. I also rerendered the png files so they look a bit less pixely now.

I can make any other adjustments as necessary.


certik:
This is a big improvement, so I am going to merge it.

certik:
Can you also add a new version of https://github.com/fortran-lang/fortran-lang.github.io/blob/b1a92bee30a6b0454bc6ece006c1e1f8822a853d/favicon.ico ? That is what appears in the Tab in a browser and it would greatly help to have the white F instead of a transparent F for better visibility.

jacobwilliams:
Yes I can do that later today.

jacobwilliams:
see: #15

PR #16, opened 2020-04-22
Update: compilers page with Flang information.

### LINKED ISSUE #6: F18 information outdated
It seems that the [F18](https://github.com/flang-compiler/f18) front-end branch has been accepted as the brand new [flang](https://github.com/llvm/llvm-project/tree/master/flang/) and is part of the official LLVM project. The former flang compiler still exists [here](https://github.com/flang-compiler/flang).

I can take a shot at updating the page in the upcoming days, if no one else does it first.

LKedward:
Fixes #6

### DIFF
--- a/compilers.md
+++ b/compilers.md
@@ -21,10 +21,14 @@ with gfortran.
 [LFortran](https://lfortran.org) is a modern, interactive, LLVM-based Fortran
 compiler.
 
-### LLVM Flang (formerly F18)
+### Flang
 
-[Flang](https://github.com/llvm/llvm-project/tree/master/flang/#flang) is an open source compiler 
-and part of the LLVM project.
+[Flang](https://github.com/llvm/llvm-project/tree/master/flang/#flang) is an
+open source compiler and part of the LLVM project (formerly
+called [f18](https://github.com/flang-compiler/f18)).
+
+[Legacy Flang](https://github.com/flang-compiler/flang) is an open source
+compiler that targets LLVM, originally based on the PGI compiler frontend.
 
 ## Commercial compilers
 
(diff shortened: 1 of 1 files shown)


milancurcic:
This looks good to me. @certik would you add anything to the description?

ivan-pi:
I should point out, that it seems like only the F18 parser is complete. So to get an executable you would still have to use the "old" flang.

LKedward:
Good point. I'll update to re-include link to "old" flang.

certik:
@gklimowicz, is this a good wording for the Flang compilers?

We are trying list all Fortran compilers at https://fortran-lang.org/compilers/.

Regarding the order, for open source compiler we should order based on the number of GitHub stars, so that should be this order:

Flang https://github.com/llvm/llvm-project (4400)
GFortran https://github.com/gcc-mirror/gcc (3700)
LFortran https://github.com/lfortran/lfortran (131)

Yes, it's unfair to LFortran, as both Flang and GFortran are benefiting from being part of a large compiler collection, but I am ok with that for now.

Regarding the commercial compilers, we should list them in the order of their relative usage, which we don't know, but we can do our best to estimate. (We plan to add all the commercial compilers eventually.)

certik:
This was fixed in #24.

gklimowicz:
@certik Just as a comment after this issue was closed, I would suggest that, like the commercial compilers, you list the open source compilers in order of presumed use. That would suggest (for now), GFortran, Current Flang (it's Arm's commercial compiler), LFortran and new Flang.

certik:
@gklimowicz I've already done that here: #25. I listed both Flangs in front of LFortran (we can re-evaluate in a year). I agree that GFortran should be first, as I have done in #25.

PR #18, opened 2020-04-23
Add explore page for listing community projects.

LKedward:
Inspired by your [comment](https://github.com/fortran-lang/stdlib/issues/110#issue-551592218) @certik, I've put together a very simple and (hopefully) user-friendly community project index page.

The 'Development' top-level page has been replaced with an 'Explore' page which lists foremost the _stdlib_ and _fpm_ projects followed by a searchable index of community code packages also broken down by category.

Packages are listed centrally by metadata in a yaml file, everything else is generated.
I've populated a few popular packages to start off with.
Eventually this could involve some integration with FPM.

Search functionality is fairly rudimentary based on substrings.

I've included automated badges for github stars _etc._ as done [here](https://maulingmonkey.com/guide/cpp-vs-rust/#crates-of-note), though do note they take some time to load due to the API - there might be a way to cache them, but I can't think how exactly.

Feedback welcome.

__Preview this pull request statically [here](https://lkedward.github.io/fortran-lang.github.io/pr/explore-code/)__

### DIFF
--- a/_data/curated_projects.yml
+++ b/_data/curated_projects.yml
@@ -17,54 +17,412 @@
 #  'version'     (optional)
 #
 # Browsable category names (lowercase!):
-#   'numerical', 'libraries', 'programming', 'scientific'
+#   'numerical', 'libraries', 'scientific', 'examples', 'programming'
+#   'graphics', 'data-types', 'io', 'interfaces', 'strings'
 #
 # When adding a new library:
 #  - Check it shows up in the expected category page(s)
 #  - Temporarily use the category 'preview' and check 
 #     check the badges render correctly at /explore/category/preview
 #
+# No need to repeat keywords from 'description' in 'tags'
+#
+#
 
-# --- Libraries / Programming ---
+# --- General Libraries ---
 
 - name: functional-fortran
   github: wavebitscientific/functional-fortran
   description: Functional programming for modern Fortran
-  categories: libraries programming
+  categories: libraries
   tags: functional filter fold map
-
-- name: erloff
-  url: https://gitlab.com/everythingfunctional/erloff
-  description: Errors and logging for fortran
-  categories: libraries programming
-  tags: errors logging
-  license: "BSD 3-Clause"
-
-- name: json-fortran
-  github: jacobwilliams/json-fortran
-  description: A Fortran 2008 JSON API
-  categories: libraries programming
-  tags: json io
-  license: none
-
-- name: gtk-fortran
-  github: vmagnin/gtk-fortran
-  description: A cross-platform library to build Graphical User Interfaces (GUI)
-  categories: libraries programming
-  tags: gui gtk graphical user interface
 
 - name: fortran-utils
   github: certik/fortran-utils
   description: Various utilities for Fortran programs
-  categories: libraries programming
-  tags: constants types sorting io mesh spline ppm hdf5 lapack
+  categories: libraries io
+  tags: constants types sorting mesh spline ppm hdf5 lapack
   version: none
 
 - name: Open Coarrays
   github: sourceryinstitute/OpenCoarrays
   description: A parallel application binary interface for Fortran 2018 compilers.
-  categories: libraries programming
+  categories: libraries
   tags: mpi openshmem gfortran
+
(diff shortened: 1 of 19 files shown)


certik:
This is great, thank you so much @LKedward !

We need to ensure all the codes from https://github.com/fortran-lang/stdlib/wiki/List-of-popular-open-source-Fortran-projects are there.

It's great that it's in the yaml format, that way we can write scripts to sort them by the number of stars.

milancurcic:
I love this, thank you Laurence for taking the initiative.

I have a few high-level suggestions for now:

* **Page and tab name**: "Explore" seems a little vague to me. I expected that it meant "Explore Fortran language" which would be more like a tutorial. Should we use a more specific word here, perhaps "Packages"?
* **Fortran-lang organisation projects**: I'd like to de-empasize "organization" and emphasize community. I suggest rewording this header as "Fortran-lang community projects", and adding this website and repo as the 3rd project. I'd like us to encourage people to discover this website as an editable and contributable project in motion, rather than a static place on the internet.
* **Featured community projects**: Considering point above, then we should reword this to avoid confusion with this specific community (fortran-lang.org) and because not all projects listed are community projects. Perhaps a broader header, **Featured open source projects** is more appropriate.
  - **Category sub-pages**: I recommend not navigating to a level 3 sub-page for each category, but to have one longer page (level 2) that lists all categories and projects, with a hyperlinked category TOC at the top. Then the visitor could easily browse all categories and packages without navigating away. I personally start to get lost when websites have level 3 sub-page.
  - **Specific categories**: I don't think these categories work well. Specifically, they're all broad and have overlap between each other. I suggest having more specific and narrowly defined categories, like Linear Algebra, Simulation, Graphics, General purpose, and perhaps others.

The rewording suggestions are probably easy to implement here. If you agree, you can include them as part of this PR.

For others I suggest we open specific follow-up PRs to improve each part of the page.

certik:
This will really help the Fortran community. Later on some of these become part of fpm registry, etc. But for now this is what is needed.

LKedward:
Thanks for the comments!
@milancurcic I agree with you re rewording - I'll update with your suggestions.

>   * **Category sub-pages**: I recommend not navigating to a level 3 sub-page for each category, but to have one longer page (level 2) that lists all categories and projects, with a hyperlinked category TOC at the top. Then the visitor could easily browse all categories and packages without navigating away. I personally start to get lost when websites have level 3 sub-page.

Unfortunately I disagree here. I personally find long scrolling pages difficult to navigate efficiently and therefore poor design. I think your point can be easily remedied using appropriate signposting (_e.g._ [breadcrumbs](https://en.wikipedia.org/wiki/Breadcrumb_navigation#Websites); I will attempt to demonstrate this better here. We can discuss this in a separate issue for the site as a whole?

>   * **Specific categories**: I don't think these categories work well. Specifically, they're all broad and have overlap between each other. I suggest having more specific and narrowly defined categories, like Linear Algebra, Simulation, Graphics, General purpose, and perhaps others.

Yep this is true - as I've added more packages, it's become clearer which categories to use and they are now more distinct. The category system is very extensible, so future PRs can refine easily.

> This is great, thank you so much @LKedward !
> 
> We need to ensure all the codes from https://github.com/fortran-lang/stdlib/wiki/List-of-popular-open-source-Fortran-projects are there.

Thanks @certik, I've been through that whole list now; for this initial PR I've prioritised adding libraries and modules that would be of use to Fortran programmers over large domain-specific scientific applications. Having been through the list (and as you note also) I find that the number of stars is actually a poor indicator of package utility IMHO. I find they are much more like social media 'likes' - we see this in the skew towards large scientific applications, which are domain-specific and hence of limited utility to the wider Fortran community. We can discuss criteria for indexing packages in another issue?

milancurcic:
This is good to go, I will merge. Great addition to the site!

We can discuss and refine any specifics in separate issues and PRs.

LKedward:
Hi @milancurcic , I didn't realise this had come out of draft status - I haven't pushed the final commits from my local repo!

milancurcic:
Sorry, my bad, I misunderstood!

PR #19, opened 2020-04-24
Fortran Packages

LKedward:
@milancurcic  These are the final edits for #18  that I didn't push. Sorry!


milancurcic:
Excellent, I love the styling, thank you. 

The categories of packages look great. You're right about using sub-pages over one long page--I didn't realize how many packages there would be. And these aren't even all of them.

milancurcic:
@certik take a look at the [preview](https://lkedward.github.io/fortran-lang.github.io/pr/explore-code/).

certik:
This is really awesome. The GitHub info (stars, license, issues, ...) take a long time to load for me (but that's good enough for now).

+1 to merge as is.

We should also setup criteria how to include a package there --- right now the criterion I think was at least 29 stars at GitHub and Fortran as the majority language (as determined by GitHub). I would be for lowering the number of stars, for example if we lower to 15, then one of my other 100% Fortran project would qualify: https://github.com/certik/hfsolver  :), what is nice about this one is that it has all kinds of very good computational routines that could be useful for others. But if we lower the number of stars, then I want to do it in a systematic way, I want to see every single GitHub project included. I think we achieved it with the > 29 stars.

PR #24, opened 2020-04-24
Suggest some tweaks for Ondřej.

gklimowicz:
@certik I've written some proposed changes for F18 and current Flang.

This seemed like the easiest way to suggest the wording change.

Let me know if you would like something else.


certik:
Thank you @gklimowicz, I appreciate it. This is perfect, and if you are ok with the wording, then so are we.

PR #27, opened 2020-04-24
Add more commercial Fortran compilers

certik:
And describe each of them using their own words as copied from their
websites.

### DIFF
--- a/compilers.md
+++ b/compilers.md
@@ -41,16 +41,70 @@ based on the NVIDIA/PGI commercial compiler.
 ### Intel
 
 [Intel Parallel Studio XE](https://software.intel.com/en-us/parallel-studio-xe)
-is Intel's suite of compilers, tools, and libraries for Fortran, C, C++, and 
-Python. 
-Intel offers the Linux edition of their compiler suite for free for
+is Intel's suite of compilers, tools, and libraries for Fortran, C, C++, and
+Python. Intel offers the Linux edition of their compiler suite for free for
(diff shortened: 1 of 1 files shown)


compilers.md
```diff
@@ -41,16 +41,70 @@ based on the NVIDIA/PGI commercial compiler.
...
-### IBM
+### NAG
 
-[IBM XL Fortran Compiler](https://www.ibm.com/us-en/marketplace/xl-fortran-linux-compiler-power) is available for Linux on POWER9 architectures.
+The latest [NAG Fortran Compiler](https://www.nag.com/nag-compiler)
+release (7.0) has extensive support for legacy and modern Fortran features including parallel programming with coarrays, as well as additional support for programming with OpenMP.
 
-### NAG
+The Compiler also provides significant support for Fortran 2018 (atomic
+operations, events and tasks, plus other smaller features), almost all of
```
milancurcic, line 54: What are tasks? Do you mean teams?

certik: I just copied what was at their page.

certik: I would suggest to just keep it. It's exactly their wording, so our job should be to just update it if they change their website, or they tell us what wording to use. Otherwise we should assume their website is correct.

certik:
I merged it. Now the compiler page looks decent, and we can keep improving it with further PRs.

compilers.md
milancurcic: Sounds good. Could be a NAG extension.

PR #45, opened 2020-04-30
Add the rest of the Fortran compilers

certik:
Every compiler from the following pages should now be listed (either
active or discontinued):

https://en.wikipedia.org/wiki/List_of_compilers#Fortran_compilers

http://fortranwiki.org/fortran/show/Compilers

https://fortran.bcs.org/2017/fortran_2003_2008_compiler_support.pdf


certik:
Available now at https://fortran-lang.org/compilers/

We are now in a pretty good shape. Turns out there are still 12 commercial compilers that seem maintained (some more than others). That's impressive.

PR #46, opened 2020-05-02
Github actions build & PR preview

### LINKED ISSUE #41: Create a Preview for each PR
Right now we have to test a PR locally that it works. Our CI should create a preview of each PR, so that we can simply see it right away. Typically that is achieved by pushing the PR into some review github pages as a branch with a url prefix so each PR has its own prefix and we can check that the page looks good.

LKedward:
This PR aims to resolve #41.

The approach here is to use github actions to explicitly build the site on push as opposed to the automatic github Jekyll build used currently; this action builds the site using Jekyll on the master branch and pushes to the 'gh-pages' branch. (Repo settings will therefore need to be changed by admin here to point to 'gh-pages' branch).

Additional actions are added for pull requests which are built and pushed to the 'gh-pages' branch under the path 'pr/<pr_number>'; therefore pull requests can be previewed at fortran-lang.org/pr/46/ for example.

To enable the 'pr/<pr_number>' sub-path, the site source has been updated such that all internal site links and references are prepended with `{{site.baseurl}}`, see [here](https://byparker.com/blog/2014/clearing-up-confusion-around-baseurl/) for explanation.

I'm not an expert on github actions so if anyone has more experience and feedback then let me know!

This has been tested on a private repo, but I'm not sure if any extra steps are required for this organisation repo.

### DIFF
--- a/.github/workflows/buildPRPreview.yml
+++ b/.github/workflows/buildPRPreview.yml
@@ -53,6 +53,7 @@ jobs:
         grep baseurl _config.yml|grep -v "#" |awk -F : '{print $2}' | xargs > base_url
         sed -i '/baseurl:/d' _config.yml
         echo "baseurl: $(cat base_url)/${{env.BUILD_DIR}}" >> _config.yml
+        echo "pull_request: ${{github.event.number}}" >> _config.yml
 
     # Run Jekyll build and copy output to PUBLISH_DIR
     - name: Build Jekyll Site
--- a/.github/workflows/buildPRPreview.yml
+++ b/.github/workflows/buildPRPreview.yml
@@ -1,0 +1,80 @@
+# Github action to build Jekyll site for a pull request preview
+#
+#  Built site is pushed to 'gh-pages' branch
+#   under directory pr/pr_id/
+#
+
+name: Jekyll PR Preview
+
+on:
+  pull_request:
+    branches: [ master ]
+
+env:
+  SRC_DIR: src
+  PUBLISH_DIR: gh-pages
+  SITE_URL: https://lkedward.github.io/gh-actions-test
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    
+    steps:
+
+    # Checkout PR branch into SRC_DIR
+    - name: Checkout pr/${{github.event.number}}
+      uses: actions/checkout@v2
+      with:
+        path: ${{env.SRC_DIR}}
+    
+    # Checkout existing gh-pages branch into PUBLISH_DIR
+    - name: Checkout gh-pages
+      uses: actions/checkout@v2
+      with:
+        path: ${{env.PUBLISH_DIR}}
+        ref: 'gh-pages' 
+    
+    # Setup Jekyll in SRC_DIR
+    - name: Setup Jekyll
+      run: |
+        cd ${{env.SRC_DIR}}
+        sudo gem install bundler
+        bundle config path .bundle
+        bundle install
+    
+    # Sets Jekyll output directory using PR id
+    - name: Define output directory
+      run: echo "::set-env name=BUILD_DIR::pr/${{github.event.number}}"
+
+    # Appends PR-based output directory to Jekyll baseurl variable
+    - name: Update Jekyll site config
+      run: |
+        cd ${{env.SRC_DIR}}
+        grep baseurl _config.yml|grep -v "#" |awk -F : '{print $2}' | xargs > base_url
+        sed -i '/baseurl:/d' _config.yml
+        echo "baseurl: $(cat base_url)/${{env.BUILD_DIR}}" >> _config.yml
+
+    # Run Jekyll build and copy output to PUBLISH_DIR
+    - name: Build Jekyll Site
+      run: |
+        cd ${{env.SRC_DIR}}
+        bundle exec jekyll build -d _site/${BUILD_DIR}
+        cp -r _site/* ../${{env.PUBLISH_DIR}}/
+        touch ../${{env.PUBLISH_DIR}}/.nojekyll
+
+    # Commit and push changes to remote/gh-pages
+    - name: Commit and push to gh-pages
+      uses: EndBug/add-and-commit@master
+      with:
+        cwd: ${{env.PUBLISH_DIR}}
+        message: "Jekyll build, preview ${{env.BUILD_DIR}} (on:pull_request)"
+        ref: 'gh-pages'
+      env:  
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    # Write comment to PR with preview link
+    - name: Comment on PR with preview link
+      uses: thollander/actions-comment-pull-request@master
+      with:
+        message: "This PR has been built with Jekyll and can be previewed at: <${{env.SITE_URL}}/${{env.BUILD_DIR}}/>"
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/buildPRPreview.yml
+++ b/.github/workflows/buildPRPreview.yml
@@ -13,7 +13,7 @@ on:
 env:
   SRC_DIR: src
   PUBLISH_DIR: gh-pages
-  SITE_URL: https://lkedward.github.io/gh-actions-test
+  SITE_URL: https://fortran-lang.org
 
 jobs:
   build:
--- a/.github/workflows/buildPRPreview.yml
+++ b/.github/workflows/buildPRPreview.yml
@@ -18,6 +18,7 @@ env:
 jobs:
   build:
     runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.body,'#no_preview') == false
     
     steps:
 
--- a/.github/workflows/buildSite.yml
+++ b/.github/workflows/buildSite.yml
@@ -1,0 +1,59 @@
+# Github action to build Jekyll site and commit to gh-pages branch
+#
+#  Built site is pushed to 'gh-pages' branch
+#
+
+name: Jekyll Build
+
+on:
+  push:
+    branches: [ master ]
+
+env:
+  SRC_DIR: src
+  PUBLISH_DIR: gh-pages
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    
+    steps:
+
+    # Checkout master branch into SRC_DIR
+    - name: Checkout master
+      uses: actions/checkout@v2
+      with:
+        path: ${{env.SRC_DIR}}
+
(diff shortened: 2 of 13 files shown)


certik:
Thanks for this! Yes, that's exactly what we need. How do I test it out?

LKedward:
> ... How do I test it out?

I'm going to see if I can merge into the master of my fork and demonstrate a working build and PR.

milancurcic:
Thank you @LKedward! Once you nail down the workflow, please document each step from a developer's point of view, either in a "How to contribute" section in README.md, or in CONTRIBUTING.md. I also don't know how this works.

LKedward:
> Thank you @LKedward! Once you nail down the workflow, please document each step from a developer's point of view, either in a "How to contribute" section in README.md, or in CONTRIBUTING.md. I also don't know how this works.

Ah yes good point, I'll make sure to document this.

LKedward:
To test this out I have merged into my personal fork:

- This is the site built using [this github action](https://github.com/LKedward/fortran-lang.github.io/runs/640144116?check_suite_focus=true):
  <https://lkedward.github.io/fortran-lang.github.io/>

 - This is a pull request preview built with [this github action](https://github.com/LKedward/fortran-lang.github.io/runs/640145207?check_suite_focus=true):
  <https://lkedward.github.io/fortran-lang.github.io/pr/3/>
   - for this pull request:  <https://github.com/LKedward/fortran-lang.github.io/pull/3>

The rendered site files (for both the normal site and for PRs) are pushed to the 'gh-pages' orphan branch <https://github.com/LKedward/fortran-lang.github.io/tree/gh-pages>.

__Before this can be merged__, a 'gh-pages' branch needs to be created in this main repository by an admin as an __[orphan branch](https://gist.github.com/seanbuscay/5877413)__

milancurcic:
Sounds good, I'll go ahead and create the orphan branch.

certik:
@milancurcic if you have time to set this up, please go ahead! I am quite busy today.

milancurcic:
@LKedward done. Is this PR ready to merge?

LKedward:
> @LKedward done. Is this PR ready to merge?

Cheers @milancurcic, yep I'm happy with it - ready to merge!

milancurcic:
Okay, just merged. Do we now need to have a dummy PR to trigger the build? Notice that the site is currently broken.

LKedward:
> Okay, just merged. Do we now need to have a dummy PR to trigger the build? Notice that the site is currently broken.

No, it's been [built and pushed](https://github.com/fortran-lang/fortran-lang.github.io/runs/640910737?check_suite_focus=true) okay.

Can you check in the repo settings that the github pages are pointing to 'gh-pages' branch.
Like this:
Screenshot from 2020-05-03 19-08-44

LKedward:
Oh no, I think we can't use the gh-pages branch on github.io repos!
I didn't take this into account here. I think we need to revert since I can't think a solution atm.

milancurcic:
I just realized it too:

Screenshot from 2020-05-03 14-08-33

milancurcic:
Okay I just reverted this PR.

Regarding a possible solution, could we:

1. Rename this repository so that it's not a user page, for example fortran-lang-website or similar;
2. Set it up to serve from gh-pages

LKedward:
> Okay I just reverted this PR.

Thanks.
Okay it looks like we have to do 1 but keep this repo:
So the current repo contents is moved to a separate repository and the github actions push the built html to this repository.
It's a bit of an upheaval unfortunately.

LKedward:
> It's a bit of an upheaval unfortunately.
By this I just mean that all contributors will have to checkout a different repository.

Regarding the github action workflows I've put together here: there is very little that needs changing, I only need to change the repository that it pushes to.

certik:
That was a mistake of using a repo and serving it directly. A better approach is to have a repository with the sources, and the CI then builds the page from master and pushes it somewhere else to be served.

Second, the name of this repository fortran-lang.github.io is misleading anyway because it is actually served at fortran-lang.org. So how about we rename this repository to fortran-lang.org, to keep all the history, stars, issues, etc. 

Then we create a new repository that would actually serve fortran-lang.org. So that it is not confusing, perhaps it can be called fortran-lang.org.deploy or something like that.

certik:
Now when I think about it, let's just keep https://github.com/fortran-lang/fortran-lang.github.io as is, if GitHub allows it, but it will become the source repository. The fortran-lang.org domain will be served from the new repository, say fortran-lang.org.deploy. The same setup from my previous comment, but we don't need to rename this repository, so people don't need to update their setup.

LKedward:
> That was a mistake of using a repo and serving it directly. A better approach is to have a repository with the sources, and the CI then builds the page from master and pushes it somewhere else to be served.

I disagree here, this approach, of using the 'gh-pages' branch is quite standard for so-called _project_ sites, see [here](https://help.github.com/en/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#choosing-a-publishing-source). Unfortunately I didn't realise github doesn't allow this approach for so-called _organisation_ sites. (I can see no actual reason why it is disallowed here)

> Second, the name of this repository fortran-lang.github.io is misleading anyway because it is actually serves at fortran-lang.org. So how about we rename this repository to fortran-lang.org, to keep all the history, stars, issues, etc.
> 
> Then we create a new repository that would actually serve fortran-lang.org. So that it is not confusing, perhaps it can be called fortran-lang.org.deploy or something like that.
> 
> And then setup a CI at fortran-lang.org that checks each PR (it builds it and pushes somewhere to be looked at, not at fortran-lang.org), and then every master will get built and pushed into fortran-lang.org.deploy to be served at fortran-lang.org.

__We have to keep the rendered html at fortran-lang.github.io__ since this is a so-called _organisation_ site which serves the site at `https://<organization>.github.io`. If we use any other repository it will not be an _organisation_ site but simply a _project_ site which is not served at the site root but under `https://<organization>.github.io/<repository>`.

See here for full explanation:
<https://help.github.com/en/github/working-with-github-pages/about-github-pages#types-of-github-pages-sites>

My proposition is similarly based on a two repository approach, but the unbuilt source (current repo contents) is moved to a new repository and fortran-lang.github.io is used for deployment via CI for the reasons mentioned above.

Unfortunately as I said, and as you point out, it is an upheaval meaning stars, issues etc are effectively lost.

milancurcic:
> That was a mistake of using a repo and serving it directly.

No, I did it on purpose as an org page so that repository pages like stdlib automatically inherit the top-level domain. 

I'm okay with either approach. For the site in production repository name, just fortran-lang.org is my preference.

LKedward:
To lay out my proposed solution more formally:

- Rename existing `fortran-lang.github.io` repo to `fortran-lang-source`
  (We get to keep stars, issues, history, etc)

- Create new `fortran-lang.github.io` repo

- Update pull request CI to deploy to master on new `fortran-lang.github.io`

certik:
@LKedward that's fine, this will work. 

Note that I routinely deploy from a deploy repository, e.g. https://www.theoretical-physics.com/, is served by https://github.com/certik/tfc-deploy and pushed into it from https://gitlab.com/certik/theoretical-physics/. But perhaps it's different if we are part of an organization? I don't know.

LKedward:
The only reason the original workflow didn't work here is because github allows the workflow for project sites but not for organisation/user sites. (I had the workflow working in my fork, but because it was a fork it was a project site not an organisation site.)

milancurcic:
We don't have to use an organization page. I chose that because of 2 benefits:

1. You don't need to maintain a production repo (the one being served), it's done automatically and under the hood by GitHub;
2. Project pages inherit top-level domain.

I don't think there are any other advantages. The obvious downside is that you can't do what we're now trying to do here, and we want to do it.

If the repo is called <orgname>.github.io **and** you chose to serve a page from it, then it's an org page (and ditto for <username>).

However, this doesn't mean that fortran-lang.github.io (this repo) must be an org page. If you disable serving in the settings, it's not a page anymore. 

There's a number of solutions here that preserve repo metadata (stars, issues, etc.). 

My favorite actually is to:

1. Rename this repo so it's not an org page (preserves metadata);
2. Deploy to gh-pages branch. 

But deploying to master of another repo is okay too. It just seems like unnecessary added complexity of carrying a separate repo for deployment.

milancurcic:
> We have to keep the rendered html at fortran-lang.github.io since this is a so-called organisation site which serves the site at https://<organization>.github.io. If we use any other repository it will not be an organisation site but simply a project site which is not served at the site root but under `https://<organization>.github.io/<repository>`.

@LKedward I don't think this matters at all because GitHub will redirect `https://<organization>.github.io/<repository>` to fortran-lang.org anyway. Am I wrong?

certik:
It's a clean approach to just have a repository with the sources. But I am fine either way.

What do you mean by inherit top level domain? I think one can always serve to subdomains such as docs.fortran-lang.org with different repositories.

milancurcic:
> What do you mean by inherit top level domain?  I think one can always serve to subdomains such as docs.fortran-lang.org with different repositories.

Example: https://fortran-lang.org/stdlib/. That page is served from the fortran-lang/stdlib repo. It's just a convenience feature, nothing major. 

Yes, we can easily serve other project pages as subdomains, it's just a matter of adding a CNAME record.

I defer the choice of final implementation for this to you and Laurence. What matters to me is that we have automatic previews and preserve metadata, for both of which it looks like we have solutions.

certik:
I see. I think we should not use this master domain feature. Rather, I would serve the whole webpage as a single generated page from a single repo. I feel that's cleaner. We can still pull out docs from other repos.

LKedward:
> 
> @LKedward I don't think this matters at all because GitHub will redirect `https://<organization>.github.io/<repository>` to fortran-lang.org anyway. Am I wrong?

Ah okay, yep you are right here. 

> My favorite actually is to:
> 
> 1. Rename this repo so it's not an org page (preserves metadata);
> 2. Deploy to gh-pages branch.
> 

I agree this is by far the simplest and least-disruptive solution.
We get to keep metadata and this pull request can be merged as is.

Does anyone know how this will affect existing forks?

LKedward:
I've reopened this PR in #54 for when the repo has been renamed.
I think `fortran-lang.org` makes sense as a repo name, what do you think?

milancurcic:
I agree. @certik what do you think?

certik:
Yes.

milancurcic:
Okay, I'm gonna rename the repo now.

milancurcic:
The rename seems like it worked. The website had about 1 minute of downtime while I was applying the custom domain in the settings. It is now served from gh-pages. I think we can now move forward and merge #54

LKedward:
Sounds good!

PR #69, opened 2020-05-12
Arpack: fix version and license

certik:


certik:
Weird, the website now says version not found, but there are releases for Arpack.

LKedward:
> Weird, the website now says version not found, but there are releases for Arpack.

Not sure why this is happening - I'll look into it.

PR #71, opened 2020-05-12
Remove OpenBLAS

certik:
The package does not satisfy our criteria, in particular:

Relevance

* the package must be primarily implemented in Fortran
* or provide a complete Fortran interface to an existing package
* or be purposed solely towards software development in Fortran.

None of these apply. They do not provide a Fortran interface to another
package, and OpenBLAS is not purposed solely towards software
development in Fortran. Regarding the first bullet point, the package is
primarily developed in assembly, all new code is in assembly. Fortran is
not the primary implementation language, despite technically having more
lines than their implementation language (assembly).
They have a lot of Fortran code (in fact the majority of lines) because
they have several copies of LAPACK. They keep applying fixes from
reference LAPACK (which is already included in our index, so we should
not be including it twice). OpenBLAS is in the same category as SciPy.
They copied a mature Fortran project / code, but use a different
language to further develop the package (assembly and Python
respectively). In the SciPy's case, the Python code now has more lines
than Fortran, in OpenBLAS case the assembly code does not yet have more
lines, but eventually it will.

Uniqueness

* the package shall not be a fork or minor revision of existing packages

As far as Fortran is concerned, this package is just a fork of LAPACK.
There is no new Fortran code that isn't in LAPACK that other projects
could use.

----

We can think (see #72) if we should have categories for projects that happen to
have Fortran code included, even it is not further developed. OpenBLAS
and SciPy would be two major libraries to include there. Until then, we
should remove OpenBLAS from our listing.

### DIFF
--- a/_data/package_index.yml
+++ b/_data/package_index.yml
@@ -441,12 +441,6 @@
 
 # --- Numerical libraries ---
 
-- name: OpenBLAS
-  github: xianyi/OpenBLAS
-  description: Optimized BLAS library based on GotoBLAS2
-  categories: numerical
-  tags: blas linear algebra
-
 - name: LAPACK
   github: xianyi/OpenBLAS
   description: Routines for numerical linear algebra


milancurcic:
Indeed OpenBLAS doesn't expose Fortran bindings to its capability. I had assumed it did, but never checked. Fortran code is truly just a dependency here. If we made Fortran bindings to say, libcurl, which would then be a dependency, we wouldn't call the bindings a C project.

I agree with removing it.

certik:
I think OpenBLAS *is* be exposing its functionality into Fortran using the Lapack interface. But it would be like we took libcurl, and rewrote some internal subroutines into Fortran, and only used Fortran to keep improving over libcurl --- that would not be a C project anymore.

milancurcic:
I didn't get that impression from the [documentation](https://github.com/xianyi/OpenBLAS/wiki/User-Manual). They expose a C interface to Fortran code. They don't expose a Fortran interface to their C and Assembly code.

certik:
Yes, I think we are both saying the same thing.

Furthermore even if they exposed Fortran code to their assembly, it wouldn't make it a Fortran project.

milancurcic:
> Furthermore even if they exposed Fortran code to their assembly, it wouldn't make it a Fortran project.

It wouldn't be strictly a "Fortran project", but it would qualify to list it in the registry because:

> or provide a complete Fortran interface to an existing package

If OpenBLAS had Fortran bindings to its unique value (over reference BLAS and LAPACK), it would be very important to include it.

certik:
I see, so it would be in the interfaces section.

I need to check, maybe they provide an interface.

The interfaces is a special section.

For me the main value of our Packages section is to show and promote codes that are actively developed in Fortran, to show the Fortran community that there are open source codes out there and to connect us together. And to show some success stories to motivate us.

The interfaces section is a secondary goal for me, those are useful to have and use obviously. But I don't view these as Fortran codes, although they would obviously be packaged by fpm later on.
Sort of like the issue #72 which I think is also very useful to have. But these are not Fortran packages in my understanding of the word.

LKedward:
I agree that according to our criteria, unless OpenBLAS contains explicit Fortran interfaces, it should not be listed.

> For me the main value of our Packages section is to show and promote codes that are actively developed in Fortran, to show the Fortran community that there are open source codes out there and to connect us together. And to show some success stories to motivate us. The interfaces section is a secondary goal for me, those are useful to have and use obviously. But I don't view these as Fortran codes, although they would obviously be packaged by fpm later on. Sort of like the issue #72 which I think is also very useful to have. 

__I have a different view of the Packages section:__ it is already well-accepted that many large applications (mainly scientific) are actively developed in Fortran. What is less accepted is that Fortran is a suitable language and has a sufficient ecosystem for __developing new codes__.
The Packages section shows off the large offering of Fortran libraries that are available to support code development - this is because someone looking to develop in Fortran will interested in what libraries already exist to make their life easier. As part of this, the interfaces category is absolutely essential.

To summarise my view:
- The existence of large scientific codes already written in Fortran is unlikely to heavily influence someone to choose Fortran for their new project (unless they want to link against them);

- Rather, the ecosystem (libraries, tools and community) are much more of an influencing factor;

- In the absence of an online package registry, the packages section serves this purpose.

> But these are not Fortran packages in my understanding of the word.

If they contain c-binding interfaces written in Fortran, then I consider them Fortran regardless of whether they implement functionality or not. (Creating a c-binding module can be a lot work)

certik:
OpenBLAS can be used from Fortran easily just like any other LAPACK implementation. I've done it.

It doesn't contain modern Fortran interfaces for linear algebra like we are creating in stdlib, but it nevertheless works from Fortran.

So I would say it qualifies for the interfaces section. It's a package that you can use from Fortran.

Good points. It looks like the purpose of the Package section is twofold. In the broader sense to show what packages are readily available to use in Fortran. And in the narrower sense to show that Fortran is used to develop new open source code. Everything except interfaces qualifies for the second point.

certik:
So I think I vote to move OpenBLAS to interfaces.

This is unlike SciPy, as SciPy cannot directly be used in Fortran, while OpenBLAS can.

milancurcic:
Okay, if that's the case, I'm okay with keeping it in "Interfaces". I am still concerned that I couldn't find OpenBLAS documentation on Fortran bindings, and I had assumed that if you're just calling Fortran LAPACK included OpenBLAS, you're not gaining anything over reference LAPACK. But I trust @certik and if he says that OpenBLAS has a meaningful and useful Fortran binding, I'm okay with it. 

I want to also caution about the "Interfaces" category. If you looked at our existing categories, you'll see that "Interfaces" is the only category that says "how something works" rather than "what it does". 

Similar to @LKedward, for me by far the main purpose of this page is for new developers to be able to find the tool they need for their Fortran work. Not so much for us to prove there are pure Fortran packages out there, or to be proud of it. That too is important, but is secondary to me.

In that sense, when a newcomer to Fortran arrives to the packages page, the "Interfaces" category is the only one that doesn't really tell them what's in there, only that the packages in there are bindings to other languages. However, I don't think this category is necessarily harmful, and is okay to keep it.

LKedward:
> So I think I vote to move OpenBLAS to interfaces.

Okay I see your point here, we should keep it.

> I am still concerned that I couldn't find OpenBLAS documentation on Fortran bindings, and I had assumed that if you're just calling Fortran LAPACK included OpenBLAS, you're not gaining anything over reference LAPACK. 
> 

My understanding is that all the LAPACK interfaces are purely explicit (explicit shape etc.) so they don't need explicit Fortran interfaces to be visible to the calling code.
(Presumably because LAPACK predates the iso c binding)

As to OpenBLAS usefulness, I only ever use OpenBLAS because it is significantly faster than the reference LAPACK.

certik:
As @LKedward said, OpenBLAS does not provide new functionality (as far as I know), but the main purpose of it is that it provides a drop in replacement for LAPACK that is much faster, as in, one of the fastest out there. It's a great library.

Yes, LAPACK predates iso_c_binding, and OpenBLAS just follows what LAPACK does. The tradition there is to just provide the symbols and hope that they work (but they do in practice, even if I don't recommend to do that for new projects).

So let's put it into into interfaces.

I think we are in agreement --- for you and @LKedward the main purpose is to list packages that are useful for Fortran developers to use. For me that is part of it. But equally important for me is to leverage our community and connect us to each other and to show what is out there, and showing how Fortran can be used --- because I think we absolutely have to prove that Fortran can be used for new projects and for new developments. That Fortran is not just for wrapping other languages and for legacy codes.

certik:
I am going back and forth on this. Maybe OpenBLAS can stay where it currently is, right next to LAPACK:

https://fortran-lang.org/packages/numerical

They both provide the same functionality, both can readily be used from Fortran, and OpenBLAS provides a much faster implementation.

I viewed the Package list as packages written in Fortran, but since both of you @milancurcic and @LKedward see it as packages that can be directly used from Fortran, then we don't care what the primary implementation language is.

I would say that OpenBLAS satisfies this bullet point:

* or provide a complete Fortran interface to an existing package

Since it provides the complete Fortran interface to their very fast (assembly+Fortran) implementation of LAPACK. As discussed, the interface is not `iso_c_binding` interface, but it's the old fashioned LAPACK interface. However it works.

So I am going to close this PR, I think OpenBLAS can stay where it is. And our understanding of the Packages section then must be: any package that can be used from Fortran right away.

PR #74, opened 2020-05-13
nitpicking :)

p-costa:


certik:
Thanks for noticing and fixing it @p-costa.

PR #76, opened 2020-05-14
Add the MESA code

certik:
I first heard about it here:

https://news.ycombinator.com/item?id=23160114

### DIFF
--- a/_data/package_index.yml
+++ b/_data/package_index.yml
@@ -667,6 +667,14 @@
   license: MIT
   version: 1.0.4
 
+- name: MESA
+  url: http://mesa.sourceforge.net/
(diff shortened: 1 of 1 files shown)


LKedward:
This code looks impressive: last release in March 2020, it uses modern Fortran with OpenMP and it has an informative website! This should definitely be listed!
yaml entry looks good to me :+1:

certik:
@LKedward yes, exactly. There might be a few more codes like this out there.

As discussed before, I think this code is not something that is probably super easy to use in other projects. However, I want it in to proudly show that Fortran is still used in actively maintained codes as the main language and for the community to see this and to perhaps learn some stuff how they do, and they can learn some stuff from other such projects.

PR #80, opened 2020-05-15
Fix broken getting started link from tutorial -> learn

aktech:
It seems the getting started link in the home page (https://fortran-lang.org/) above footer is broken as it still points to `tutorial/`, which was removed in this commit: https://github.com/fortran-lang/fortran-lang.org/commit/2adc5eba834fc34e90a7c0b216449db2fcecfdea#diff-94d9a081cd1256334373c8ca6fb6276c

### DIFF
--- a/index.html
+++ b/index.html
@@ -165,7 +165,7 @@ title: Home
       <h3>Write Fortran software</h3>
       <p>
       Or just write Fortran software for your research, business, or schoolwork.
-      You can learn how to get started <a href="/tutorial/">here</a>.
+      You can learn how to get started <a href="/learn/">here</a>.
       </p>
     </div>
 


index.html
```diff
@@ -165,7 +165,7 @@ <h3>Develop tools</h3>
       <h3>Write Fortran software</h3>
       <p>
       Or just write Fortran software for your research, business, or schoolwork.
-      You can learn how to get started <a href="/tutorial/">here</a>.
+      You can learn how to get started <a href="/learn/">here</a>.
```
LKedward, line 168:
```suggestion
      You can learn how to get started <a href="{{site.baseurl}}/learn/">here</a>.
```

certik:
Nice to see you here @aktech !

index.html
certik: Oops, I am really sorry I already merged this. I think we have to submit a new PR.

milancurcic:
Thank you and welcome @aktech !

index.html
aktech: That's an improvement, it should work without that as well.

LKedward: Thanks for the PR @aktech. We should keep to the habit of prepending all internal site links like this. See [here](https://github.com/fortran-lang/fortran-lang.org/blob/master/CONTRIBUTING.md#internal-site-links) in the (incomplete) contributor document.

aktech: Makes sense, thanks for the link. Will keep that in mind.

PR #81, opened 2020-05-15
Use {{site.baseurl}} instead of relative link

aktech:
As suggested here:
https://github.com/fortran-lang/fortran-lang.org/pull/80#discussion_r425971804 and
https://github.com/fortran-lang/fortran-lang.org/blob/master/CONTRIBUTING.md#internal-site-links

ping @LKedward @certik


certik:
Thanks Amit!

PR #84, opened 2020-05-18
Cray: Fix the name of CCE

certik:


certik:
This was an approved change by Cray, so I am just going to merge it.

PR #87, opened 2020-05-20
Add: Navbar social icons

### LINKED ISSUE #85: Move the Discourse link at the top
Currently one must scroll down to find the Discourse link. Let's put it at the top horizontal menu? I think we want this as the primary discussion forum, so that would be a great place for it.

LKedward:
Resolves #85.
I also removed the social links from the news sidebar (now that they're in the navbar) and replaced them with the fortran-lang twitter feed.


LKedward:
#build_preview

LKedward:
#delete_preview

PR #88, opened 2020-05-20
Fixes #73 lapack entry in package index.

### LINKED ISSUE #73: Fix LAPACK package entry
The github entry should be:
`Reference-LAPACK/lapack`

and license is BSD 3-Clause.

LKedward:
Fixes #73


certik:
Good catch.

PR #100, opened 2020-06-06
Add community section

LKedward:
This PR creates a new top-level page called 'Community' for the Fortran-lang projects in order to:
- Provide instructions & links for (new) contributors
- Link to project documentation
- List and acknowledge contributors

The PR includes json data files from <https://gitlab.com/fortran-lang/github_stats_data> which is used to generate the contributor lists and plots. On the new community page you can use the slider beneath the aggregate plot to filter contributions by a specific date range.
The contributions list can also be incorporated into newsletters with a fixed date range, which I have done for the June newsletter as an example.

Work for future PRs will be:
- including commit data into contributions;
- using github actions to update json data periodically.

As always, I value feedback on this - my focus has been on the structure and implementation, so any suggestions for the text content are especially welcome.

(Apologies for the size of this PR, it grew to more than I was initially planning to organise and cleanup the javascript files.)


LKedward:
#build_preview

jvdp1:
Great new page!

milancurcic:
This is excellent, thank you Laurence. Very nice contribution graphs. Are contributions == commits + comments?

LKedward:
> This is excellent, thank you Laurence. Very nice contribution graphs. Are contributions == commits + comments?

Thanks!
No, I should have said: currently it's just comments because that's what's in the json files currently - when Ondřej updates his stats scripts, it will include commits as well.
@certik

_posts/2020-06-01-Fortran-Newsletter-June-2020.md
```diff
@@ -125,11 +120,12 @@ For reference, you can read the [notes from the February meeting](/newsletter/20
...
+See the [the community page]({{site.baseurl}}/community) for full stats.
+
+<div id="gh-contributors" data-startdate="May 01 2020" height="400px" data-enddate="May 31 2020"></div>
```
certik, line 131: I would not modify old newsletters. Those were done and as far as I know all the information in there is accurate. I would treat old newsletters like published papers. As new updates and graphs become available, we can publish a new paper. But old papers should not be modified, unless there was a typo, or some obvious falsehood, in which case we should add an "Update", or "Correction".

However let's use the new stats for the next newsletter. It looks really good I think.

everythingfunctional:
This is really cool. I think it helps to gameify things enough to keep us all coming back and contributing. And the more we keep contributing the better it will get and the more it will grow and then turn into a virtuous cycle. Great addition.

certik:
@everythingfunctional you got the idea. We want people to feel proud and excited to be visible there.

_posts/2020-06-01-Fortran-Newsletter-June-2020.md
LKedward: Good point, agreed - I will revert that.

LKedward:
#build_preview

LKedward:
Thanks for the feedback!

Regarding the javascript library and JSON data files, I agree that these should ideally be left out of this repository - CI downloading is an option, but we also don't want the CI environment to be able to diverge from the local clones, _i.e._ we need a solution that brings dependencies into the local repos as well.

My preferred solution to this would be to use git submodules, however I'm aware that some people do not like submodules. I personally have never had any problems with submodules and the advantage here is that it minimally affects contributor workflow while also allowing dependencies in separate repos.
Does anyone have a strong preference not to use git submodules?

certik:
I think this can be merged as far as I am concerned.

I personally am not a big fan of submodules, it's an extra struggle. But I am also not against it, and I think it technically does what we need: keeping the big files out of the repository. And since you are @LKedward the main contributor to this website (as apparent from the statistics in this PR:), and you like submodules, then let's use them.

If too many people struggle with them, we can always move to another solution later.

Thanks again for this PR.

everythingfunctional:
I have good success using submodules as a dependency management strategy. I would vote for using them. As @certik said, we can move to something else later if need be.

LKedward:
Okay I will merge and open a new issue for the external dependencies.

LKedward:
#delete_preview

PR #109, opened 2020-06-19
better not write all functions again

MarDiehl:
just a small hint that I find helpful


MarDiehl:
I also thought about changing the example, but then decided this is pedagogical not bad: Learn the manual implementation and learn that there is an intrinsic. I also could not think of a good example for a vector reduction to a scalar. `min` and `max` are also intrinsic.
The example is anyways really basic, I would never make the size of the array an extra argument.

milancurcic:
Small PR with no objections, merging. Thank you, Martin!

PR #114, opened 2020-06-25
Update: preview action for future posts.

LKedward:
Build PR previews with `--future` flag to allow previewing of future newsletters.


milancurcic:
I'll merge so we can try it for the newsletter.

PR #119, opened 2020-07-08
Fix: broken contributor plots on safari.

### LINKED ISSUE #105: Contributor graph does not work on mobile
The contributor graph at https://fortran-lang.org/community/ does not work on my iPhone:

E6E4E61D-923A-47D2-9E16-151CF37915FA

I have iPhone 10 and Firefox. I also tried Safari, the same result.

LKedward:
Fixes #105.
There are still some formatting issues on mobile, but those can be addressed in a separate issue.

### DIFF
--- a/assets/js/gh-contributors.js
+++ b/assets/js/gh-contributors.js
@@ -281,7 +281,7 @@ ghContributorStats = function() {
             }
 
             var userData = userContribData.get(repo.issues[i].user);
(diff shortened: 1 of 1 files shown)


LKedward:
#build_preview
@certik, can you check on your iPhone?

certik:
I just checked both Safari and Firefox on iPhone and it works now! Thank you for fixing it @LKedward.

LKedward:
#delete_preview

PR #123, opened 2020-07-14
Add: license info to Futility package entry.

LKedward:
License info is not detected automatically for this package, so add explicitly.


certik:
+1

PR #127, opened 2020-07-27
Add Fortran Equation Parser

schoonovernumerics:


_data/package_index.yml
```diff
@@ -727,3 +727,11 @@
   categories: examples numerical
   tags: ode pde integral stochastic quadrature plotting
   version: none
+
+- name: Fortran Equation Parser
+  github: FluidNumerics/feq-parse
+  description: An equation parser class for interpreting and evaluating equations provided as strings.
+  categories: strings data-types
```
LKedward, line 734:
```suggestion
  categories: strings io
```

Since this is a singleton class (I think), I don't think it's suited to the `data types` category.
The `io` category encompasses parsing, so this is more suitable.

_data/package_index.yml
```diff
@@ -727,3 +727,11 @@
...
+  tags: equation-parser
```
LKedward, line 735:
```suggestion
  tags: equation parser
```

certik:
+1 to add. I haven't checked if the categories exist.

_data/package_index.yml
schoonovernumerics: I agree. Thanks for the suggestion

LKedward: Hi @schoonovernumerics, just checking have you committed/pushed the above change?

certik: @schoonovernumerics, please commit the change if you agree with it (click on the button that says "accept the change" or something like that). Then we can merge this.

certik:
Thanks! Merging.

PR #128, opened 2020-07-30
package_index: FortranCallGraph and Serialbox

dev-zero:
While looking for a callgraph generator I stumbled over the FortranCallGraph project (a static callgrapher), from there you can find the FortranTestGenerator (unittest framework based on Capture & Replay via Serialbox), and the serialization library Serialbox (from Swiss Meteo).
The FortranTestGenerator I'm not yet adding due to lack of documentation example code.

### DIFF
--- a/_data/package_index.yml
+++ b/_data/package_index.yml
@@ -188,6 +188,13 @@
   tags: unit testing
   license: BSD 2-clause
 
+- name: FortranCallGraph
+  github: fortesg/fortrancallgraph
+  description: Static source code analysis tool for Fortran
+  categories: programming
+  tags:
+  license: GNU GPL v3
+
 # --- Data types ---
 
 - name: Fortran template library
--- a/_data/package_index.yml
+++ b/_data/package_index.yml
@@ -393,6 +393,12 @@
   tags: 
   license: MIT
 
+- name: Serialbox
+  github: GridTools/serialbox
+  description: A serialization library and tools for C/C++, Python3 and Fortran
+  categories: io
+  tags: validation
+  license: "BSD 2 Clause"
 
 # --- Graphics ---
 


_data/package_index.yml
```diff
@@ -188,6 +188,13 @@
   tags: unit testing
   license: BSD 2-clause
 
+- name: FortranCallGraph
+  url: https://github.com/fortesg/fortrancallgraph
```
LKedward, line 192:
```suggestion
  github: fortesg/fortrancallgraph
```

_data/package_index.yml
```diff
@@ -386,6 +393,12 @@
   tags: 
   license: MIT
 
+- name: Serialbox
+  url: https://github.com/GridTools/serialbox
```
LKedward, line 397:
```suggestion
  github: GridTools/serialbox
```

_data/package_index.yml
```diff
@@ -386,6 +393,12 @@
...
+  description: A serialization library and tools for C/C++, Python3 and Fortran
+  categories: io
+  tags:
+  license: BSD
```
LKedward, line 401:
```suggestion
  license: "BSD 2 Clause"
```

_data/package_index.yml
```diff
@@ -386,6 +393,12 @@
   tags: 
   license: MIT
 
+- name: Serialbox
+  url: https://github.com/GridTools/serialbox
+  description: A serialization library and tools for C/C++, Python3 and Fortran
+  categories: io
+  tags:
```
LKedward, line 400:
```suggestion
  tags: validation
```

dev-zero:
@LKedward thanks, done

LKedward:
#build_preview

LKedward:
Packages and their badges are rendering properly 👍 .
I'm happy that these packages meet the criteria for listing so I'll merge.
#delete_preview

milancurcic:
Just so I understand for future packages, this one satisfies the relevance criterion:

> Relevance: the package must be primarily implemented in Fortran or provide a complete Fortran interface to an existing package or be purposed solely towards software development in Fortran.

only because it's "purposed solely towards software development in Fortran", correct?

I ask because when I first took a looked at the repo, I saw that it's a Python package with only target code samples in Fortran, so it gave me pause. 

So, the 3rd test in the relevance criterion basically allows any non-Fortran tool that is somehow to be used for Fortran development. Another obvious example is FORD. Am I interpreting this correctly?

LKedward:
> So, the 3rd test in the relevance criterion basically allows any non-Fortran tool that is somehow to be used for Fortran development. Another obvious example is FORD. Am I interpreting this correctly?

Yes, this is my reasoning. The third criterion was written this way precisely to allow FORD and similar utilities.
Ultimately fortran-lang should probably list development tools and utilities like FORD separately somewhere, but for now they allowed as 'packages' just as a convenience.

certik:
I am a little worried that we are opening a pandora box with this ---- it seems we should also list any Fortran related tools, such as LFortran, GFortran, Flang, all kinds of plugins for VSCode etc. 

Maybe we can have a dedicated tag for these.

LKedward:
> I am a little worried that we are opening a pandora box with this...

I agree, which is why a separate site page may be more suitable. Perhaps the _Compiler_ page becomes a more general page with compilers and development tools?

certik:
I think we should list all good related tools for Fortran somehow. Including things like debuggers and optimization tools if they work with Fortran. Also currently we only list open source tools (but for compilers we want to list the commercial ones too).

So it's just a question where. For now maybe we can have a tag or category for "tools" and list all these open source tools there.

PR #133, opened 2020-08-07
List FPM registry packages

LKedward:
A minimal modification to include packages from fpm-registry on the website.
They are listed in their own category ('fpm') as well as searchable via the existing search function.

This is a stopgap until we can clean this up and separate out fpm-registry like crates.io to provide more info.


LKedward:
#build_preview

LKedward:
Sorry, preview's not working because it depends on the new CI files

LKedward:
Thanks Ondřej! Yep I'm open to suggestions as to how we can improve this in the future

LKedward:
#delete_preview

certik:
I checked it on the website and it looks awesome! Thank you for implementing this.

PR #136, opened 2020-09-18
Small fix in the opening sentence on the compilers page

milancurcic:
We list a total of 16 compilers on the page, so the opening line should be fixed.


certik:
+1

PR #141, opened 2020-10-02
fix youtube embed

milancurcic:
There was an erroneous space character immediately before the `<iframe>` tag. Apparently this made it not render (??).


certik:
Thanks for fixing it!

PR #144, opened 2020-10-04
Add fypp to package index

awvwgk:
I wonder why this one is missing here

- GH repository: https://github.com/aradi/fypp
- documentation: https://fypp.readthedocs.io


LKedward (approved):
We will eventually separate programming tools like FORD, fypp and fprettify onto a separate tools page, but for the time-being it's good to have them somewhere on site. Thanks

milancurcic (approved):

PR #145, opened 2020-10-04
Sort package index

awvwgk:
Sorts the package index after the package names, this should make the package lists clearer.


LKedward (approved):
Perfect, thanks!

PR #146, opened 2020-10-04
Correct license identifiers

awvwgk:
`license: none` should be a no-go for the package index. The package index is of course no legal advice for licensing, but the information should be as accurate as possible. A package without license is unusable by definition (for open source projects), which is unfair to packages with permissive licensing conditions.
For a simple summary see: https://choosealicense.com/no-permission/

I went through all the packages with `license: none` and tried to identify the licenses. I also tried to correct all licenses which are not valid SPDX license identifiers. Many missing licenses were malformatted BSD or MIT licenses, but some had actually interesting licensing condition like original MIT license with changes under BSD-3-Clause and so on.

There are still some issues with the GPL-2.0, and GPL-3.0, which should have the `-only` or `-or-later` suffix. (L)GPL-2 and (L)GPL-3 is usually insufficient to identify the license of a package clearly, since the minor version is missing.

This PR is open for discussion.

### DIFF
--- a/_data/package_index.yml
+++ b/_data/package_index.yml
@@ -4,16 +4,16 @@
 #  'description' (mandatory) single line description
 #  'categories'  (mandatory)
 #  'tags'        (optional) used for searching. Don't include category name here.
-#  'license'     (optional) override automatic github license. Set to 'none' to hide.
+#  'license'     (optional) override automatic github license. Mandatory if automatic detection fails.
 #  'version'     (optional) override automatic github version. Set to 'none' to hide.
 #
 # Fields for non github projects:
 #  'name'        (mandatory)
 #  'url'         (mandatory)
+#  'license'     (mandatory)
 #  'description' (mandatory) single line description
 #  'categories'  (mandatory)
 #  'tags'        (optional) used for searching. Don't include category name here.
-#  'license'     (optional)
 #  'version'     (optional)
 #
 # Browsable category names (lowercase!):
--- a/_data/package_index.yml
+++ b/_data/package_index.yml
@@ -55,7 +55,7 @@
   description: Fortran command Line Arguments Parser
   categories: libraries
   tags: command line cli argument parser
-  license: none
+  license: GPL-3.0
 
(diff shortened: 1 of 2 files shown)


certik (approved):
Awesome, thanks a lot. I agree license "none" is a no-no.

awvwgk:
Many of @szaghi's projects have a GPL-3.0-only OR BSD-2-Clause OR BSD-3-Clause OR MIT license, which is quite lengthy for a badge (it doesn't state GPL-3.0-or-later for any package, I think). Not sure how to put those correctly without overflowing the rows.

LKedward:
Thanks @awvwgk, I agree with you however please note that `license: none` is a (misleading) programmatic identifier to disable rendering of the shields.io badge, it __does not indicate anywhere on site that any software is license-free__. 
Our package guidelines already state that only open source projects are accepted and they also require that the license field be specified if not detectable by github. I believe the large number of `license: none` entries may be, as you point out, that the licenses were non-standard formatting which meant I didn't know which identifier to use - at the time I opted to display no information rather than wrong information.
Thank you for tracking down this information.

We never made any promises regarding the use of SPDX identifiers, but I have no objection to adopting them now.

awvwgk:
> I agree with you however please note that `license: none` is a (misleading) programmatic identifier to disable rendering of the shields.io badge, it does not indicate anywhere on site that any software is license-free.

How about `license: off` which should map to a boolean false in yaml, has a much clearer indent and can easily used within liquid. Or you can have a separate field as `display-license: false` to avoid mixing the meta data with programmatic identifiers.

LKedward:
Are there still packages currently in the index where we still need this capability? If so, a separate field seems like the best option. If not, the behaviour should simply be removed. It already isn't allowed in future PRs according to the current guidelines.

awvwgk:
This sounds way better than my suggestion :+1:.

certik:
Thanks for fixing this. Looks like there is no opposition, so I am going to merge it.

PR #154, opened 2020-10-07
Add a link to fortran90.org

certik:
I plan to migrate information from there to fortran-lang.org, but until
this happens, it would be good to at least provide the link, so that
people who are new to Fortran can use the site to learn modern Fortran.


LKedward (approved):
Yes I agree, a lot of great content there.

certik:
Thanks for the review!

PR #156, opened 2020-10-23
Update building programs book

awvwgk:
I think the build-tools section of the building programs minibook is a bit short and the make example contains a few common gotchas, therefore I tried to update it.

I also started some minimal introductions to meson and CMake to explain the concept of high-level build systems in comparison to make. Since it is an introductory book I didn't went to much into details on using those, as this might be something for an intermediate or advanced guide.

Changes:

- replace hyphens by ~~endashs~~ emdashs
- use correct external link declaration
- rename the example source file to functions
- correct name of the generate module file set to `user_functions`
- extend the build-tools page with a detailed guide on make
  - explain concept of recursively defined variables which is commonly misused in make
- add a short guide on meson
- add a short guide on CMake

### DIFF
--- a/learn/building_programs/build_tools.md
+++ b/learn/building_programs/build_tools.md
@@ -253,8 +253,9 @@ implicit rules messing up our ``Makefile`` in surprising and harmful ways.
 Next, we have a configuration section where we define variables, in case you
 want to switch out your compiler, it can be easily done here.
 We also introduced the ``SRCS`` variable to hold all source files, which is
-more intuitive than specifying object files, we can easily create the object
-files by appending a ``.o`` suffix using the functions ``addsuffix``.
+more intuitive than specifying object files.
+We can easily create the object files by appending a ``.o`` suffix using the
+functions ``addsuffix``.
 The ``.PHONY`` is a special rule, which should be used for all entry points
 of your ``Makefile``, here we define two entry point, we already know *all*,
 the new *clean* rule deletes all the build artifacts again such that we indeed
@@ -313,7 +314,7 @@ You should find the expected (or maybe unexpected) printout after running ``make
 {% include note.html content="appending with ``+=`` to an undefined variable will produce a recursively expanded variable with this state being inherited for all further appending." %}
 
 While, it seems like an interesting feature to use, it tends to lead to
-surprising and expected outcomes. Usually, when defining variables like your
+surprising and unexpected outcomes. Usually, when defining variables like your
 compiler, there is little reason to actually use the recursive expansion at all.
 
 The same can easily be archived using the ``:=`` declaration:
--- a/learn/building_programs/build_tools.md
+++ b/learn/building_programs/build_tools.md
@@ -215,7 +215,7 @@ action has to be performed.
 In the end, we will have a look at a complete ``Makefile``.
 
 ```make
-# Disable all of make's implicit rules (similar to Fortran's implicit none)
+# Disable all of make's built-in rules (similar to Fortran's implicit none)
 MAKEFLAGS += --no-builtin-rules --no-builtin-variables
 # configuration
 FC := gfortran
@@ -247,7 +247,7 @@ clean:
 ```
 
 Since you are starting with ``make`` we highly recommend to always include
-the first line, like with Fortrans ``implicit none`` we do not want to have
+the first line, like with Fortran's ``implicit none`` we do not want to have
 implicit rules messing up our ``Makefile`` in surprising and harmful ways.
 
 Next, we have a configuration section where we define variables, in case you
@@ -264,15 +264,17 @@ Also, we slightly changed the build rule for the object files to account for
 appending the ``.o`` suffix instead of substituting it.
 Notice that we still need to explicitly define the interdependencies in the
 ``Makefile``. We also added a dependency for the object files on the ``Makefile``
-itself, in case you change the compiler, this will allow you to savely rebuild.
+itself, in case you change the compiler, this will allow you to safely rebuild.
 
 Now you know enough about ``make`` to use it for building small projects.
+If you plan to use ``make`` more extensively, we have compiled a few tips
+for you as well.
 
 {% capture note %}
 
 In this guide, we avoided and disabled a lot of the commonly used ``make``
 features that can be particularly troublesome if not used correctly, we highly
-recommend staying away from the buildin rules and variables if you do not feel
+recommend staying away from the builtin rules and variables if you do not feel
 confident working with ``make``, but explicitly declare all variables and rules.
 
 You will find that ``make`` is capable tool to automate short interdependent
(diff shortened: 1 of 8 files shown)


certik:
Quite frankly, why don't we replace that section with `fpm`? 

The document (I thought) is tailored for new users. New users either want to start a project from scratch (thus `fpm` I think will work already) or they wan to contribute to an already existing project, but then they have to follow whatever build system the existing project uses.

milancurcic:
> Quite frankly, why don't we replace that section with fpm?

I agree that this is the way to go. 

Although it may not be quite adequate yet (two implementations, unstable spec and behavior, unimplemented functionality in the Fortran client), this would help adopt new fpm users. New Fortran users would go straight to fpm.

I recommend that:

1. We merge this PR;
2. We work on the fpm guide
3. We put the fpm guide first, and we keep the existing guide as "Other build tools" or similar. It's useful to keep it.

awvwgk:
There is some diversity in build tools available, I think they should be discussed at least briefly to give some guidance to a beginner. Also, there are only few projects offering help or useful insights for the build systems they are using. I was motivated to look into this guide after a friend asked me about a good read with some more depth on compiling, linking and building.

awvwgk:
An fpm guide wouldn't need any of the information in this particular minibook, writing a separate book on learning Fortran with fpm is the better call in my opinion.

awvwgk:
Question: How do you feel about inline HTML in the minibooks? I found we have a CSS class for tips in

https://github.com/fortran-lang/fortran-lang.org/blob/5f9584472fdf1a691196931c60d0cac363427cde/assets/css/main.css#L463-L468

Using this seems to require inline HTML instead of the blockquote CSS:

```diff
diff --git a/learn/building_programs/build_tools.md b/learn/building_programs/build_tools.md
index 27d1a75..621ea2b 100644
--- a/learn/building_programs/build_tools.md
+++ b/learn/building_programs/build_tools.md
@@ -60,12 +60,15 @@ It performs actions following rules defined in a configuration file
 called ``Makefile`` or ``makefile``, which usually leads to compiling a program
 from the provided source code.
 
-> **Tip**
->
-> For an in-depth ``make`` tutorial lookup its info page. There is an online
-> version of this
-> <a href="https://www.gnu.org/software/make/manual/make.html" target="_blank" rel="noopener">info page</a>,
-> available.
+<div class="aside-tip">
+  <b>Tip</b>
+  <p>
+    For an in-depth <code>make</code> tutorial lookup its info page. There is an online
+    version of this
+    <a href="https://www.gnu.org/software/make/manual/make.html" target="_blank" rel="noopener">info page</a>,
+    available.
+  </p>
+</div>
 
 We will start with the basics from your clean source directory. Create and open
 the file ``Makefile``, we start with a simple rule called *all*:
```

LKedward:
> Edit: There seems to be an include HTML as well...

Yes please use the HTML includes for these asides.

> ... somewhat suboptimal construct with a liquid capture region

The capture region isn't a necessesity (see Quickstart book) - if you can think of way to improve the syntax for simplicity please do!

LKedward:
> An fpm guide wouldn't need any of the information in this particular minibook, writing a separate book on learning Fortran with fpm is the better call in my opinion.

I share this sentiment

learn/building_programs/build_tools.md
```diff
@@ -4,75 +4,437 @@ title: Build tools
...
+> ``make``.
+>
+> While ``make`` is indeed a handy tool to automate short interdependent
+> workflows and to build small projects, it should *never* be used to build
+> larger projects, like scientific programs.
+> In particular modern Fortran programs can hardly be handled by a ``make``
+> build system.
+>
+> To use ``make`` in a larger project, you cannot rely on ``make`` alone, usually
+> other tools are used to generate the ``Makefile`` completely or in parts.
```
LKedward, line 277: This section is strongly opinionated and without justification. It appears to reflect the author's experience with `make`; this may or may not be the case, but this is how it comes across in the text.

Any tool with advanced features will require further learning - warning off potential users implies that the tool is not fit for purpose, this is not something which we should imply for `make`.

I disagree with the absolute and generalised statement about the suitability of `make` for large projects. A balanced statement indicating the limitations of `make` and reasoning for using build configuration systems would suffice here IMO.

LKedward:
As someone who struggled to find adequate online material when learning `make` and other systems for Fortran (I still struggle with `cmake`), I think this is a good addition to this particular minibook. A separate minibook for fpm seems more appropriate - when ready this can then supplant this minibook as the recommended beginners' guide.

awvwgk:
@LKedward I won't get around a few capture statements, because some of the asides are several paragraphs with code blocks, but this should work for now. I'll have a look at the templates to figure out something nicer. Also, I noticed that the asides don't look as nice as the blockquotes, the spacing is somewhat off, maybe the CSS could use some adjustment as well.

learn/building_programs/build_tools.md
awvwgk: I'm probably opinionated on make, I used it for quite a while extensively and now completely dropped it from all my projects because I couldn't make it scale well. I'll remove the paragraph than.

LKedward: Thanks for rewording - looks good now. I will try to do a more extensive review soon.

awvwgk:
> As someone who struggled to find adequate online material when learning make and other systems for Fortran (I still struggle with cmake), I think this is a good addition to this particular minibook. A separate minibook for fpm seems more appropriate - when ready this can then supplant this minibook as the recommended beginners' guide.

I added a few sections on common gotchas in `make` as well. If wanted I can also expand on some of the more useful `make` features, like `vpath`, string manipulation or recursive `make`.

ivan-pi:
> As someone who struggled to find adequate online material when learning `make` and other systems for Fortran (I still struggle with `cmake`), I think this is a good addition to this particular minibook. A separate minibook for fpm seems more appropriate - when ready this can then supplant this minibook as the recommended beginners' guide.

I second these thoughts. The available resources on Cmake and make targeted at Fortran are scarce. Most of the time I resorted to just adapting the makefiles from existing Fortran projects, not knowing what exactly are they doing. Having the minibooks in one place will hopefully make it easier for beginners to progress from small projects to intermediate and large projects where a build system is needed (at least until fpm is mature).

certik:
Sounds good, I agree with the comments above.

awvwgk:
This PR has been stale for almost a month now. Is there still interest in this?

milancurcic (approved):
Thank you for the nudge, @awvwgk, I forgot about this PR. 

I reviewed it again and I like the changes.

I only suggest that we replace en dashes with em dashes. En dashes are typically used for ranges and em dashes to break up sentences, see this [en dash guide](https://www.thepunctuationguide.com/en-dash.html) and this [em dash guide](https://www.thepunctuationguide.com/em-dash.html).

Originally, I would have suggested not using spaces around em dashes, which tend to be used only in newspapers with narrow columns. This allows them to break across lines. However, I changed my mind about this because our website is mobile friendly and can have narrow columns on smaller devices. Spaces around em dash can thus improve readability on such devices.

milancurcic:
With no objections, I'll go ahead and merge this.

PR #158, opened 2020-10-24
add dnf install gcc-gfortran for Fedora 22+

jvdp1:
Since Fedora 22, `dnf` is the default package manager (instead of `yum`).
Note: the command `yum` is still availble, but is only an alias to `dnf`.


certik (approved):
Looks good to me.

PR #160, opened 2020-10-27
Add in-depth introduction for Fortran with make

awvwgk:
I just gave ``make`` and Fortran another try and realized again why I dropped it from *all* my project in the first place. Nevertheless, the attempt of writing this up might still be helpful for somebody learning ``make``.

The idea of the guide is the following, take a real world example from the Fortran package index with medium complexity (more than one source file, but not a huge project), which is not using ``make`` as build system and describe the process of porting it to ``make`` as insightful as possible.

This guide describes:
- creating simple functions
- creating maps in ``make``
- foreach loops over lists
- using eval to generate ``make`` expressions from lists
- handling of dependencies
- partly generating the ``Makefile`` with external tools
- wildcard functionality to detect source files

Let me know what you think about this.

### DIFF
--- a/learn/building_programs/project_make.md
+++ b/learn/building_programs/project_make.md
@@ -29,8 +29,8 @@ Before going into detail with ``make``, there are a few points to consider first
 If you think ``make`` is suitable for your needs, than you can start writing
 your ``Makefile``. For this course we will use real world examples from the
 package index, which use (at the point of writing) different build systems
-than ``make``. This guide should present a generally useful style to write
-``make``, but also serve as demonstration of useful features.
+than ``make``. This guide should present a general recommended style to write
+``make``, but also serve as demonstration of useful and interesting features.
 
 {% include tip.html content="Even if you find ``make`` unsuitable to build your project, it is *the* tool to automate workflows defined by files. Maybe you can leverage its power in a different context." %}
 
@@ -172,8 +172,8 @@ Some compilers also offer to generate dependencies in ``make`` format.
 
 Before diving into the dependency generation, we will outline the concept of
 a robust take on the dependency problem.
-Each source file provides (``module``) or requires (``use``) modules,
-a robust approach can process all source files independently
+First, we want an approach that can process all source files independently,
+while each source file provides (``module``) or requires (``use``) modules.
 When generating the dependencies only the name of the source file and the
 module files are known, no information on the object file names should be
 required.
--- a/learn/building_programs/project_make.md
+++ b/learn/building_programs/project_make.md
@@ -220,7 +220,7 @@ To generate the respective dependency map automatically we will use an
 #!/usr/bin/awk -f
 
 BEGIN {
-    # Fortran is case insensitive, disable case sensitivity for matching:
+    # Fortran is case insensitive, disable case sensitivity for matching
     IGNORECASE = 1
 }
 
@@ -246,7 +246,7 @@ $2 ~ /^[a-zA-Z][a-zA-Z1-9_]*,?$/ {
     gsub(/,/, "", $2)
     # count used module names per file to avoid using modules twice in our list
     if (usec[FILENAME,$2]++ == 0) {
-        # add to the module list, the generated module name is expected
+        # add to the used modules, the generated module name is expected
         # to be lowercase, the FILENAME is the current source file
         use[++iu] = sprintf("$(%s) += $(%s.mod)", FILENAME, tolower($2))
     }
@@ -257,9 +257,9 @@ $2 ~ /^[a-zA-Z][a-zA-Z1-9_]*,?$/ {
 # - the second argument ($2) can be everything, as long as delimited by quotes
 $1 ~ /^(#:?)?include$/ &&
 $2 ~ /^["'].+["']$/ {
-    # Remove a quotes from the included file name
+    # Remove quotes from the included file name
     gsub(/'|"/, "", $2)
-    # count included files per file to avoid having duplicated in our list
+    # count included files per file to avoid having duplicates in our list
     if (incc[FILENAME,$2]++ == 0) {
         # Add the included file to our list, this might be case-sensitive
         inc[++ii] = sprintf("$(%s) += %s", FILENAME, $2)
@@ -275,13 +275,17 @@ END {
 }
 ```
 
-This script makes a few assumptions about the source code it parses, so it will not work with all Fortran code, but for this example it will suffice.
+This script makes a few assumptions about the source code it parses, so it will
+not work with all Fortran code (namely submodules are not supported), but for
+this example it will suffice.
 
 {% capture note %}
 
 The above script uses the ``awk`` language, which is designed for the purpose
-of text stream processing. In ``awk`` you can define groups which are evaluated
-on certain events, *e.g.* when a line matches a specific pattern.
+of text stream processing and uses a C-like syntax. In ``awk`` you can define
+groups which are evaluated on certain events, *e.g.* when a line matches a
+specific pattern, usually expressed by a
+<a href="https://en.wikipedia.org/wiki/Regular_expression" target="_blank" rel="noopener"> regular expression</a>.
 
 This ``awk`` script defines five groups, two of them use the special pattern
 ``BEGIN`` and ``END`` which are run before the script starts and after the script
--- a/learn/building_programs/project_make.md
+++ b/learn/building_programs/project_make.md
@@ -220,32 +220,54 @@ To generate the respective dependency map automatically we will use an
 #!/usr/bin/awk -f
 
 BEGIN {
+    # Fortran is case insensitive, disable case sensitivity for matching:
     IGNORECASE = 1
 }
 
+# Match a module statement
+# - the first argument ($1) should be the whole word module
+# - the second argument ($2) should be a valid module name
 $1 ~ /^module$/ &&
 $2 ~ /^[a-zA-Z][a-zA-Z1-9_]*$/ {
+    # count module names per file to avoid having modules twice in our list
     if (modc[FILENAME,$2]++ == 0) {
+        # add to the module list, the generated module name is expected
+        # to be lowercase, the FILENAME is the current source file
         mod[++im] = sprintf("%s.mod = $(%s)", tolower($2), FILENAME)
     }
 }
 
+# Match a use statement
+# - the first argument ($1) should be the whole word use
+# - the second argument ($2) should be a valid module name
 $1 ~ /^use$/ &&
 $2 ~ /^[a-zA-Z][a-zA-Z1-9_]*,?$/ {
+    # Remove a trailing comma from an optional only statement
     gsub(/,/, "", $2)
+    # count used module names per file to avoid using modules twice in our list
     if (usec[FILENAME,$2]++ == 0) {
+        # add to the module list, the generated module name is expected
+        # to be lowercase, the FILENAME is the current source file
         use[++iu] = sprintf("$(%s) += $(%s.mod)", FILENAME, tolower($2))
     }
 }
 
+# Match an include statement
+# - the first argument ($1) should be the whole word include
+# - the second argument ($2) can be everything, as long as delimited by quotes
 $1 ~ /^(#:?)?include$/ &&
 $2 ~ /^["'].+["']$/ {
+    # Remove a quotes from the included file name
     gsub(/'|"/, "", $2)
+    # count included files per file to avoid having duplicated in our list
     if (incc[FILENAME,$2]++ == 0) {
+        # Add the included file to our list, this might be case-sensitive
         inc[++ii] = sprintf("$(%s) += %s", FILENAME, $2)
     }
 }
 
+# Finally, produce the output for make, loop over all modules, use statements
+# and include statements, empty lists are ignored in awk
 END {
     for (i in mod) print mod[i]
     for (i in use) print use[i]
--- a/learn/building_programs/project_make.md
+++ b/learn/building_programs/project_make.md
@@ -1,0 +1,452 @@
+---
+layout: book
+title: An introduction to make
+permalink: /learn/building_programs/project_make
+---
+
+We have shortly discussed the basics of ``make``, this chapter gives thoughts
+and strategies to make ``make`` scale for larger projects.
+
+Before going into detail with ``make``, there are a few points to consider first:
+1. ``make`` is a Unix tool and might give you a hard time when porting to non-Unix
+   platforms. That said, there are also different flavors of ``make`` available,
+   not all might support the features you want to use.
+2. while ``make`` gives you full control over the build process, it also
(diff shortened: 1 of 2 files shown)


arjenmarkus:
What is the level of sophistication you intend to reach? I have seen incredibly complicated Makefiles (mostly generated via tools like automake) that are not intended to be read or understood by mere mortals, but I have never encountered functions, maps, loops over lists or eval, at least not in Makefiles that were intended for mere mortals. 
And I do not think it is necessary to make it so sophisticated ;). But I may be mistaken. I would say that describing the last three points in depth would suffice for most if not all Fortran projects, but you can easily prove me wrong. After all, an essential tool like make has been described in preciously few books at any length, or I have missed most of them.

awvwgk:
The aim is to show a concise, scalable and still readable handwritten Makefile. The resulting Makefile will probably be sophisticated, but not complicated or bloated like an autogenerated Makefile. The part I wrote so far targets only the generation of dependencies in an automatic way, explaining the other features while showing use-cases for them.

milancurcic:
I like this idea and think it will be useful. It will be useful for me for sure. I even own a book on GNU Make that I struggled learning anything useful from.

If the guide shows to be advanced, we can always give a heads-up note at the beginning of the guide, stating the intended level and who it's for.

certik:
Here is an example of a manual Makefile that we use in stdlib:

https://github.com/fortran-lang/stdlib/blob/3733ae3624ec6852f9fe624d1f91fbd0423bdf01/src/Makefile.manual

and it has module dependencies, and yet is seems simpler than the one proposed in this PR (no `awk`, no `addsuffix`, no `patsubst`, ...). Wouldn't it make sense to recommend the simpler approach that we did in `stdlib`?

awvwgk:
There is certainly more than one style of ``make`` possible, the question is which we want to advertise.

I find the substitution reference mechanism of ``make`` rather obscure, even if it avoids writing out `patsubst` or `addsuffix` functions explicitly. So technically the two Makefiles are setup quite similar setup, except for the stdlib Makefile is using recursively expanded variables while the guide is using normal variables instead.

certik:
I see. The approach that I like the most is `fpm`, and let `fpm` generate any other build system (if so desired), so even though I think I prefer the stdlib's style of makefile, I am fine with your approach to makefiles also. We can put your style in, and later if people have time, they can submit the other approach also.

awvwgk:
@certik The scope of this PR is to give an advanced introduction to ``make``. It depends on the status of #156 and I'll adjust this PR to account for any review comments on ``make`` style there. I opened it as draft to see if there is any interest to have an advanced ``make`` guide in the minibooks at all and to discuss if the format I have chosen seems appropriate.

> The approach that I like the most is fpm, and let fpm generate any other build system (if so desired)

I agree that ``make`` would be my least favourite choice as build system for Fortran, since there are plenty other build systems like CMake, meson, fpm, ... which are easier to use. But I don't think that letting fpm generate a Makefile removes the need of an introduction to ``make``. An fpm generated Makefile would be on the same level as an autotools or CMake generated one and not supposed to be read or modified by the user (at least in my opinion). Also, I would prefer to discuss fpm related features at the fpm repository instead.

awvwgk:
Let's give this PR another bump, since it has been stale for a while now.

This guide builds on the basic `make` introduction from #156, which is now part of the building programs book. It presents an intermediate to advanced overview over some powerful functionality and less known concepts of `make`. The idea is to show some interesting and unusual approaches how to actually use `make` to solve problems, going beyond the stuff you will usually find in a random `Makefile`.

The automation of the dependencies is one of the main issues faced when using `make` for Fortran projects, of course one can hard code them in the `Makefile` and avoid the topic, but this is in my opinion not a good answer on how to deal with module dependencies in `make`.

If this is not suited for the building programs book, I would suggest to start a new book on build systems and development tools as intermediate minibook instead.

awvwgk:
Let's give this PR another bump, since it has been stale for a while now.

This guide builds on the basic `make` introduction from #156, which is now part of the building programs book. It presents an intermediate to advanced overview over some powerful functionality and less known concepts of `make`. The idea is to show some interesting and unusual approaches how to actually use `make` to solve problems, going beyond the stuff you will usually find in a random `Makefile`.

The automation of the dependencies is one of the main issues faced when using `make` for Fortran projects, of course one can hard code them in the `Makefile` and avoid the topic, but this is in my opinion not a good answer on how to deal with module dependencies in `make`.

If this is not suited for the building programs book, I would suggest to start a new book on build systems and development tools as intermediate minibook instead.

certik:
I just talked with @awvwgk over video regarding this issue and I am fine with merging it as is. While we all work hard on `fpm`, it will still take some time before it can be used with larger production codes, and even when it can be used, not everybody will be using it for historical or other reasons. And thus having an authoritative source of "best practices" for Make and CMake with Fortran would always be helpful, even if they might not be the first approach we would recommend to beginners down the road.

So I am for merging this, and see what the response is, and if we get feedback, we can iterate on it.

@milancurcic, are you ok to merge as is?

certik (approved):

milancurcic:
I will review it over the weekend and will have minor edit suggestions. But overall I think it's a great addition. Thank you for putting it together!

arjenmarkus:
I agree - it is a straightforward receipe for creating makefiles in a
generic way. And it definitely contains features that I was unaware of.

Op vr 18 dec. 2020 om 20:33 schreef Milan Curcic <notifications@github.com>:

awvwgk:
Might be a good occasion to check if the #build_preview is working again.

learn/building_programs/project_make.md
```diff
@@ -0,0 +1,452 @@
...
+   maintaining your ``Makefile``, this is usually time you cannot use for
+   developing your source.
+3. you can work with your ``Makefile``, but think about the other developers
+   on your project (your friends or your coworkers or maybe somebody visiting your
+   open source project). How much time do you expect them to spend learning your
+   ``Makefile`` and would they be able to debug or add functionalities?
+4. pure ``make`` will not scale. You will soon write or use auxiliary programs
+   to dynamically or statically generate your ``Makefile``, those introduce
+   dependencies and possible sources of errors. Testing and documenting those
+   tools should not be neglected.
```
milancurcic, line 27:
```suggestion
4. Pure ``make`` will not scale. You will soon add auxiliary programs
   to dynamically or statically generate your ``Makefile``. Those introduce
   dependencies and possible sources of errors. The effort needed to test and document those
   tools should not be underestimated.
```

learn/building_programs/project_make.md
```diff
@@ -0,0 +1,452 @@
...
+
+If you think ``make`` is suitable for your needs, than you can start writing
+your ``Makefile``. For this course we will use real world examples from the
+package index, which use (at the point of writing) different build systems
```
milancurcic, line 31:
```suggestion
package index, which (at the time of writing) use build systems other
```

learn/building_programs/project_make.md
```diff
@@ -0,0 +1,452 @@
...
+``make``, but also serve as demonstration of useful and interesting features.
+
+{% include tip.html content="Even if you find ``make`` unsuitable to build your project, it is *the* tool to automate workflows defined by files. Maybe you can leverage its power in a different context." %}
+
+
+## Getting started
+
+For this part we will work with
+<a href="https://github.com/jacobwilliams/fortran-csv-module/tree/1.2.0" target="_blank" rel="noopener"> the Fortran CSV module (v1.2.0)</a>.
+Aim is it to write a ``Makefile`` to compile this project to a static library.
```
milancurcic, line 42:
```suggestion
Our goal is to write a ``Makefile`` to compile this project to a static library.
```

learn/building_programs/project_make.md
```diff
@@ -0,0 +1,452 @@
...
+Start by cloning the repository
+
+```
+git clone https://github.com/jacobwilliams/fortran-csv-module -b 1.2.0
+cd fortran-csv-module
+```
+
+{% include note.html content="For this part we will work with the code from tag ``1.2.0``, to make it as reproducible as possible. Feel free to use the latest version or another project instead." %}
+
+The project just downloaded uses FoBiS as build system, you can check the
```
milancurcic, line 52:
```suggestion
This project uses FoBiS as build system, and you can check the
```

learn/building_programs/project_make.md
```diff
@@ -0,0 +1,452 @@
...
+        ├── csv_kinds.f90
+        ├── csv_module.F90
+        ├── csv_parameters.f90
+        ├── csv_utilities.f90
+        └── tests
+            ├── csv_read_test.f90
+            ├── csv_test.f90
+            └── csv_write_test.f90
+
+We find seven differently named Fortran source files, the four in ``src`` should
```
milancurcic, line 74:
```suggestion
We find seven different Fortran source files; the four in ``src`` should
```

learn/building_programs/project_make.md
```diff
@@ -0,0 +1,452 @@
...
+be compiled and added to a static library while the three in ``src/tests``
+contain individual programs depending on this static library.
```
milancurcic, line 76:
```suggestion
contain individual programs that depend on this static library.
```

learn/building_programs/project_make.md
```diff
@@ -0,0 +1,452 @@
...
+Also, right now the the source files and dependencies are specified explicitly,
+which results in several additional lines even for such a simple project.
+
+
+## Automatically generated dependencies
+
+The main drawback of ``make`` for Fortran is the missing capability to
+determine module dependencies. This is usually solved by either adding those
+by hand or automatically scanning the source code with an external tool.
+Some compilers also offer to generate dependencies in ``make`` format.
```
milancurcic, line 171: I didn't know this. Which compilers can do this?

Perhaps we can add a few examples in parentheses, i.e. "Some compilers (e.g. this compiler or that compiler)"

learn/building_programs/project_make.md
```diff
@@ -0,0 +1,452 @@
...
+
+Before diving into the dependency generation, we will outline the concept of
+a robust take on the dependency problem.
+First, we want an approach that can process all source files independently,
+while each source file provides (``module``) or requires (``use``) modules.
+When generating the dependencies only the name of the source file and the
+module files are known, no information on the object file names should be
```
milancurcic, line 178:
```suggestion
module files are known, and no information on the object file names should be
```

learn/building_programs/project_make.md
```diff
@@ -0,0 +1,452 @@
...
+$(src/csv_utilities.f90): $(csv_parameters.mod)
+$(src/tests/csv_read_test.f90): $(csv_module.mod)
+$(src/tests/csv_test.f90): $(csv_module.mod)
+$(src/tests/csv_write_test.f90): $(csv_module.mod)
+```
+
+The same strategy of creating a map is already used for the module files, now
+it is just expanded to the object files as well.
+
+To generate the respective dependency map automatically we will use a simple
```
milancurcic, line 219:
```suggestion
To generate the respective dependency map automatically we will use an
```

For somebody not familiar with awk, this script is anything but simple. I'd even say it looks daunting. For such a reader it could be discouraging to read the words "simple script" and then see the thing that follows.

learn/building_programs/project_make.md
```diff
@@ -0,0 +1,452 @@
...
+END {
+    for (i in mod) print mod[i]
+    for (i in use) print use[i]
+    for (i in inc) print inc[i]
+}
+```
+
+This is a particular simple script which makes a few assumptions on the source
+code it is going to parse, so it will certainly not work for all Fortran code,
+but for this example it will suffice.
```
milancurcic, line 261:
```suggestion
This script makes a few assumptions about the source code it parses, so it will not work with all Fortran code, but for this example it will suffice.
```

learn/building_programs/project_make.md
```diff
@@ -0,0 +1,452 @@
...
+    $(src/tests/csv_test.f90) += $(iso_fortran_env.mod)
+    $(src/csv_parameters.f90) += $(csv_kinds.mod)
+    $(src/csv_module.F90) += $(csv_utilities.mod)
+    $(src/csv_module.F90) += $(csv_kinds.mod)
+    $(src/csv_module.F90) += $(csv_parameters.mod)
+    $(src/csv_module.F90) += $(iso_fortran_env.mod)
+
+Note that the scripts output will use recursively expanded variables and not
+define any dependencies yet, because out-of-order declaration of variables
+might be necessary and do not want to create any target by accident.
```
milancurcic, line 331:
```suggestion
might be necessary and we do not want to create any target by accident.
```

milancurcic (approved):
I left many minor editorial suggestions. They mostly address sentences that are too long or some redundant wording. 

I'm not sure about the whole awk scripting for dependencies. As you write, it doesn't generally work, only for some simpler cases. Being unfamiliar with awk, reading this part was daunting to me. When I saw the script and read that it won't generally work but only for simpler cases, I was already decided that I'll be writing out my dependencies by hand. So it wasn't very useful for me but may be for others. I guess my only problem with this part is that the tutorial spends a significant fraction teaching how to use a non-Fortran low-level tool to do something that will work only sometimes. At the same time, I think it's clear from the text that "you don't have to do this if you don't want to", so the reader can decide for themselves.

Otherwise I think it's good.

learn/building_programs/project_make.md
awvwgk: `ifort -gen-deps` is one example, about I never used it so far.

awvwgk:
```suggestion
Some compilers (like the Intel Fortran compiler) also offer to generate dependencies in ``make`` format.
```

awvwgk:
The `awk` script is a weak spot, it only shortcoming is that it cannot deal with submodules, but otherwise it can handle most Fortran `module` and `use` statements, which are not using line breaks and/or whitespace hacks in fixed format. The difficulty is that you actually require a full Fortran parser to solve the dependency generation issue generally.

Hardcoding dependencies is a maintenance effort I found unacceptable for using `make` in Fortran projects, especially for projects I develop together with others. The only robust solution I found so far was over-engineering the `Makefile` generation or moving to another build system.

awvwgk:
I added some comments on how `awk` is processing the Fortran source to extract the information we are looking for.

ivan-pi:
If you have by chance the O'Reilly book *Unix for FORTRAN Programmer* written by Mike Loukides it contains a nicely written chapter on `make`.  Since the book was first published in 1990 the content is limited to F77, where `make` might have been sufficient to resolve all dependencies. If interested I can send you an excerpt from the book.

awvwgk:
@ivan-pi Thanks for the offer. My best source of knowledge on Unix tools was/is the built-in documentation so far. For example the info page of `make` is excellent (check `info make` or https://www.gnu.org/software/make/manual/make.html).

awvwgk:
Thanks everybody for the comments and suggestions. With two approvals I'll go ahead and merge this PR.

PR #164, opened 2020-11-03
Test Twitter CI

milancurcic:


milancurcic:
#tweet This is a test tweet.

milancurcic:
@p-costa I double-checked the names of the secrets (look okay to me), and I updated their values. I also double-checked that we have r+w access for the app in the Twitter account. We still get the same error.

This confuses me though:

tweet_error

What is `consumer-key`? Is that a different name for one of the secrets or something different?

Do you see if these are incorrect in any way?

secrets

And here are the permissions.

twitter_app_permissions

What do you think?

milancurcic:
I just looked in [ethomson/send-tweet-action](https://github.com/ethomson/send-tweet-action). Here's the snippet from their README:

```
name: Send a Tweet
on: [push]
jobs:
  tweet:
    runs-on: ubuntu-latest
    steps:
      - uses: ethomson/send-tweet-action@v1
        with:
          status: "Hi, this is a test!"
          consumer-key: ${{ secrets.TWITTER_CONSUMER_API_KEY }}
          consumer-secret: ${{ secrets.TWITTER_CONSUMER_API_SECRET }}
          access-token: ${{ secrets.TWITTER_ACCESS_TOKEN }}
          access-token-secret: ${{ secrets.TWITTER_ACCESS_TOKEN_SECRET }}
```

Note the secrets names here. They are different from what [twitter-together](https://github.com/gr2m/twitter-together/blob/master/docs/01-create-twitter-app.md) instructs.

Should we rename `TWITTER_API_KEY` to `TWITTER_CONSUMER_API_KEY` and `TWITTER_API_SECRET` to `TWITTER_CONSUMER_API_SECRET`?

p-costa:
Hi Milan,

It is no problem that they are different as long as the names in our CI are consistent with those in the github settings:

```
        status: ${{steps.extract-tweet.outputs.body}}
        consumer-key: ${{secrets.TWITTER_API_KEY}}
        consumer-secret: ${{secrets.TWITTER_API_SECRET_KEY}}
        access-token: ${{secrets.TWITTER_ACCESS_TOKEN}}
        access-token-secret: ${{secrets.TWITTER_ACCESS_TOKEN_SECRET}}
```
I tested it on my own twitter account. (what matters is they are passed correctly to send-tweet-action). So it all looks good to me... Can it be that the keys are swapped? I can dig further into this sometime tomorrow.

p-costa:
Perhaps trying the classical solution removing those secrets and adding them again? It seems that the secrets are not passing correctly to the CI. We get an error from the first secret the CI tries to fetch from the repo.

milancurcic:
It's possible that the keys are incorrect but I find it unlikely considering I did it twice. The error message "consumer-key is required input" just doesn't make it sound like the key is incorrect, but missing altogether.  It's possible that the org secrets are not correctly propagated to the repo secrets, even though they say they are. I will try this next.

PR #170, opened 2020-11-17
Fix PSBLAS github address mistake

ivan-pi:
Fixes the typo I made.


ivan-pi:
Since @certik and @LKedward already reviewed the package is suitable and this is only a tiny correction, I will go ahead and merge.

PR #171, opened 2020-12-01
Add Sebastian to the list of authors in the building-programs mini-book

milancurcic:
I have merged #156 but only later realized that we didn't add @awvwgk to the list of authors. I did so in this PR. @awvwgk are you okay with this?


awvwgk (approved):
Sure, I didn't really thought about this part when writing it up.

PR #178, opened 2020-12-07
Fix build preview

awvwgk:
- use new syntax to set environment variables

### DIFF
--- a/.github/workflows/buildPRPreview.yml
+++ b/.github/workflows/buildPRPreview.yml
@@ -48,7 +48,7 @@ jobs:
     
     # Sets Jekyll output directory using PR id
     - name: Define output directory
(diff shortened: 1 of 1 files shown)


awvwgk:
#build_preview

awvwgk:
This is still using the workflow file from the main repo rather than from this branch. Seems like the event triggering the build preview through a comment is not associated with the PR but with the latest commit on the main branch instead?

So you have to trust me that this will fix the issue with the build preview :wink:

LKedward (approved):
Thanks @awvwgk, this all looks good 👍

milancurcic (approved):
Thank you!

urbanjost:
#build preview

Issue #68, opened 2020-05-12
Add packages from List-of-popular-open-source-Fortran-projects


certik:
Go over:

https://github.com/fortran-lang/stdlib/wiki/List-of-popular-open-source-Fortran-projects

and add the rest of the packages.

LKedward:
- [x] OpenBLAS
- [x] fortran-machine
- [x] LAPACK
- [x] ElmerFEM
- [x] WRF
- [x] fds
- [x] functional-fortran
- [x] Quantum ESPRESSO
- [x] fluidity
- [x] fortranlib
- [x] json-fortran
- [x] Nek5000
- [x] cp2k
- [x] fortran2018-examples
- [x] NASTRAN-95
- [ ] specfem3d
- [ ] NWChem
- [x] gtk-fortran
- [ ] CFL3D
- [x] fortran-utils
- [x] SHTOOLS
- [x] OpenCoarrays
- [x] arpack-ng
- [ ] DFTB+
- [ ] NASTRAN-93
- [ ] MPAS
- [x] neural-fortran
- [x] NetCDF-Fortran
- [ ] CTSM
- [x] Castro
- [x] ABINIT
- [x] FLAP
- [x] bspline-fortran
- [x] FOODIE
- [x] forpy
- [x] OFF
- [ ] WAVEWATCH III
- [ ] MOM6
- [x] VTKFortran
- [x] datetime-fortran
- [ ] node-fortran
- [ ] Fortran-Astrodynamics-Toolkit
- [x] pyplot-fortran
- [ ] Starlink
- [ ] WPS
- [x] ogpf
- [x] pFUnit
- [x] tcp-client-server
- [x] ftl
- [x] fgsl
- [ ] OpenCMISS
- [x] StringiFor
- [ ] wrf_hydro_nwm_public
- [x] SciFortran
- [ ] bandup
- [x] LaGriT
- [x] stdlib
- [x] clfortran
- [x] fox
- [ ] HANDE-QMC
- [x] FEconv
- [ ] lesgo
- [x] FortranPatterns
- [ ] FV3-GFS
- [x] coretran
- [ ] tsunami
- [x] fdict
- [ ] CFD
- [ ] ICAR
- [ ] MODFLOW6
- [ ] PIC Skeleton Codes
- [ ] IAMR
- [ ] ADflow
- [ ] PoisFFT
- [ ] sigma
- [ ] aenet
- [x] DBCSR
- [x] GALAHAD
[...]

PR #186, opened 2020-12-19
Add missing packages from list of popular Fortran projects

awvwgk:
This should add *all* the remaining packages from https://github.com/fortran-lang/fortran-lang.org/issues/68 which have an open source license.

*Important:* since my background is electronic structure theory, I'm mostly not familiar with weather modeling or astrodynamics, which many of those projects are targeting. Somebody with experience in this fields should have a look over the descriptions and add tags as appropriate.

This PR will eventually close https://github.com/fortran-lang/fortran-lang.org/issues/68.

### DIFF
--- a/_data/package_index.yml
+++ b/_data/package_index.yml
@@ -979,6 +979,12 @@
   categories: scientific
   version: none
 
+- name: MITgcm
+  github: MITgcm/MITgcm
+  description: M.I.T General Circulation Model master code and documentation repository
+  categories: scientific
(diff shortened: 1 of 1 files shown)


LKedward (approved):
Many thanks @awvwgk, this is great! All looks good to me :+1:

awvwgk:
Let's try the #build_preview again.

certik (approved):
I think this looks great, thank you!

awvwgk:
#delete_preview

Issue #204, opened 2021-02-10
Mention comp.lang.fortran and Stack Overflow under "looking for help"


Beliavsky:
I suggest that links to [comp.lang.fortran](https://groups.google.com/g/comp.lang.fortran), [Stack Overflow: Fortran](https://stackoverflow.com/questions/tagged/fortran), and [Reddit: Fortran](https://www.reddit.com/r/fortran/) be added at https://fortran-lang.org/learn/ .

awvwgk:
Close this due to discussion #219.

PR #219, opened 2021-03-05
Add links to other related Fortran communities

awvwgk:
- r/fortran
- comp.lang.fortran
- Stack Overflow

Closes https://github.com/fortran-lang/fortran-lang.org/issues/204

### DIFF
--- a/_data/learning.yml
+++ b/_data/learning.yml
@@ -64,6 +64,18 @@ reference-links:
     url: https://www.fortran90.org/
     description: Fortran Best Practices guide, Python/Fortran Rosetta Stone, Fortran FAQ
 
+  - name: comp.lang.fortran
+    url: https://groups.google.com/g/comp.lang.fortran
+    description: Mailing list for all Fortran related things
+
+  - name: "Stack Overflow: Fortran"
+    url: https://stackoverflow.com/questions/tagged/fortran
+    description: Question and answers to coding questions
+
+  - name: r/fortran
+    url: https://www.reddit.com/r/fortran/
+    description: Reddit for Fortran related topics
+
   - name: "Fortran 2018 Standard Interpretation Document"
     url: https://j3-fortran.org/doc/year/18/18-007r1.pdf
     description: "J3/18-007r1 F2018, specification of the base Fortran 2018 language"


LKedward:
Thanks @awvwgk and @Beliavsky for suggesting — I think adding stack overflow is a good idea since it is a natural place for people to go for help and the Fortran tag could do with more people answering questions. 

For the reasons I mention [here](https://fortran-lang.discourse.group/t/do-we-need-a-discourse-when-we-have-a-mailing-list/14/8?u=lkedward) I am hesitant to add links to the CLF and Reddit forums but I am not completely opposed as long as it is clear that the Discourse is the recommended community for seeking help.

awvwgk:
I don't really understand this concern. While @fortran-lang has the more active community, we don't have a reason to discourage participating in other communities which are not our own, do we? (putting the argument purposely to an extreme)

LKedward:
> ... we don't have a reason to discourage participating in other communities which are not our own, do we?

To expand on my Discourse comment, my main concern regards beginners and those who are new to Fortran. Consider such a beginner who, looking for help, goes to CLF with their query but is met with curt unaccommodating responses, as is common there, and thus leaves with a negative impression of the Fortran community and motivation to not continue pursuing Fortran.
With Reddit, which is more likely to attract young people, the Fortran community is not so active and so answers can vary in helpfulness, accuracy and accommodation. This again influences initial experiences for Fortran beginners.

By placing links on our site there is an implicit endorsement, which IMHO isn't deserved by either CLF or Reddit as recommended Fortran communities.

With this reflection I'd revise my previous comment to say that I am opposed to CLF and Reddit being listed.

awvwgk:
Thanks for taking a clear point here, @LKedward. I haven't participated in either r/fortran or CLF up to now, so I can't judge those communities in any meaningful way. I fully agree that we cannot promote communities that are unaccommodating or hostile to beginners or do not meet our community standards.

While I agree that the perception of Fortran for beginners is important, I want to highlight that the tagline of the learn page reads *Learning resources for beginners and experts alike* and we include in the same section expert literature like the 2018 standard interpretation document which could also scare off beginners. So the scope of the resources is broader than the perception generated for Fortran beginners.

milancurcic:
I don't support linking to CLF. I have posted there a few times and have had some positive interactions, but otherwise it's not moderated and toxic discussions and spam are not uncommon. That was one of the motivators for starting Fortran Discourse and mailing list.

I haven't been active on Stack Overflow for several years now. I've had negative experience there and have read toxic interactions. But it is moderated. I don't know how it is now so I can't recommend either way.

r/fortran I know almost nothing about.

So, in summary, I'm against linking to CLF and undecided about Stack Overflow and Reddit. But I trust @LKedward's opinion on this.

awvwgk:
Thanks everybody for the advice on this, in conclusion I'll withdraw this patch.

certik:
I agree with @milancurcic and @LKedward. I think I speak for others when I say that we are all for linking to other communities and resources, e.g. at https://fortran-lang.org/learn/, but it is important to be welcoming and nice to new users and contributors, as well as to old contributors. 

I think the Stack Overflow might be pretty nice, I had pretty good experience there, so I would be fine with linking it. I didn't have a good experience at c.l.f, and know several people (in person) who got so discouraged that they won't post there again. So that is unacceptable and I want to endorse communities where this does not happen, or very rarely.

PR #228, opened 2021-03-14
Update news.xml

Mr-codehunter:


LKedward:
Thanks for the patch @Mr-codehunter, however for the reasons specified [here](https://stackoverflow.com/a/30761004) I do not believe this change is necessary or even correct. Please let me know your reasons if you disagree, and I'll reopen if needed.

PR #233, opened 2021-03-24
Update package_index.yml

alinelena:
adding dl_poly_4 to the list of packages.

### DIFF
--- a/_data/package_index.yml
+++ b/_data/package_index.yml
@@ -1143,6 +1143,13 @@
   categories: scientific
   tags: electronic-structure wavelets density-functional-theory
   license: custom
+  
+- name: DL_POLY_4
+  gitlab: ccp5/dl-poly
+  description: classical molecular dynamics 
+  categories: scientific
+  tags: molecular-modeling
+  license: LGPL-3.0
 
 # --- Examples / demos / templates ---
 


certik (changes requested):
Here is the criteria that I am using to judge the PR: https://github.com/fortran-lang/fortran-lang.org/blob/master/PACKAGES.md

The package (https://gitlab.com/ccp5/dl-poly/) had 3 stars on GitLab, so I gave it a star. :)
I think it would satisfy the "star" requirement, as GitLab projects typically get less stars than GitHub ones.

However I think the README requirement is not satisfied:

> README: the package shall have some form of README or landing-page clearly stating the package purpose and functionality. This should also contain information on the package dependencies and the steps required to build and run.

What is the purpose of the package and the functionality? It currently does not say anything. Only from this PR I can see that it is a "classical MD". 

@alinelena would you mind please adding some brief paragraph into the README for your package to explain what it is, what features it has (if it is has some features that other classical MD packages don't have or if it is meant for teaching / learning), does it run in parallel, etc.?

Also, you should remove the space between `[Build instructions] (./building.md)` to make it a link in your README.

alinelena:
@certik I can fix almost all the things you suggest... but the star stuff sorry I do not like to entertain that, it is really teenage stuff.

certik:
@alinelena if you fix what I suggested, we will be happy to include your package. As I suggested, you do not have to worry about the number of stars. If you have a better idea how we could formulate the criteria, please let us know, we are open to suggestions.

certik:
The README now gives an overview about the project:

https://gitlab.com/ccp5/dl-poly/-/blob/88a0c62d1cda8f50703c3d088396aba9e2c0a1e2/README.md

Moreover, it lists publications in CPC which describe details about the original code as well as recent publications.

certik:
@alinelena would you mind please reopening this PR so that we can merge it?

certik:
New PR was sent at #233.

PR #254, opened 2021-05-08
Fix some links in May 2021 newsletter

zmoon:
I noticed the first of these when reading the newsletter. I did a brief pass looking for others though and found the other two.


certik (approved):
Thank you!

certik:
I think these were obvious mistakes. Thanks for spotting it and fixing it!

PR #255, opened 2021-05-08
Quickstart edits

zmoon:
I'm doing a bit of an audit of the *Quickstart Fortran Tutorial*. So far I mostly corrected some English stuff and tried to make the code style consistent within each page (but not across all pages).

I would like to make style consistent across the whole tutorial though, but that requires deciding on some things that are currently not consistent among all pages:
* code style
  - indentation (2 or 4 spaces? or 3 or other?)
  - whether to indent module `contains` rt. `module`, similarly for type `contains`
  - when to put a space after `,` (`do` loop, arguments, indexing, etc.)
  - vertical aligning of `::` and inline comments. Important?
  - should the base indentation level in a given markdown code block always be 0 (perhaps making it easier to copy), or should it be > 0 if the code would be normally indented if put into context (as done in some of the examples)?
  - always capitalize first word in comments?
* text style
  - when to use **strong** vs *emphasis* vs `code`
  - title case vs sentence case for the titles of the pages

I also have some content questions:
* what is meant by "`do concurrent` is not a basic feature of Fortran" in *Operators and Control Flow*?
* would a page on formatted printing and such fit here?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -262,6 +262,27 @@ do i = 1, 10
 ! ...
 ```
 
+Other situations besides simple indexings where white space can be omitted:
+* Aliasing in imports 
+  ```fortran
+  use, intrinsic :: iso_c_binding, only: sp=>c_float, dp=>c_double
+  ```
+* String concatentation
+  ```fortran
+  print *, 'hello '//'world'
+  ```
+* Accessing components (attributes) of derived types
+  ```fortran
+  p%x
+  p%calc_something(a, b)
+  ```
+* Around `=` when passing keyword arguments
+  ```fortran
+  call sr(a, b, c=3)
+  point = t_point(x=1., y=2.)
+  character(len=:), allocatable :: s
+  ```
+
 Capitalize the first letter for inline comments except for trailing inline comments that only consist of one word or a short phrase.
 ```fortran
 ! Compute new values
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -257,7 +257,7 @@ b2(long_i_name, long_j_name)
 b3(i + 2, j)
 call some_subroutine(a, b, an_option=.false.)
 c = [1, 2, 3, 10]
-d = [i, i = 1, 10]
+d = [(i, i = 1, 10)]
 do i = 1, 10
 ! ...
 ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -303,3 +303,6 @@ Make use of the admonition/aside [includes](_includes) (*note*, *tip*, *importan
 * *note*: extra information, something that might appear in a footnote
 * *tip*: information about best practices, practical tips
 * *important*: warnings, things to avoid, etc.
+
+Prefer including the [Oxford comma](https://en.wikipedia.org/wiki/Serial_comma). It usually makes things more clear.
+> Fortran is fast, fun, and famed.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -216,8 +216,8 @@ contains
 
   subroutine func(self)
     class(a_t), intent(in) :: self
-    if (val > 0) then
-      print *, val
+    if (self%val > 0) then
+      print *, self%val
     end if
   end function func
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -239,13 +239,13 @@ module <module name>
 end module <module name>
 ```
 
-Add spacing in expressions where it helps with readability, but otherwise limit.
+Feel free to omit spacing in expressions where it helps with readability, but generally include whitespace around operators.
 ```fortran
-y1 = a*b
-y2 = a*b + c*d
+y1 = a * b
+y2 = a*b + c*d  ! instead of a * b + c * d
 y3 = a**2 + 1
 y4 = (a*b + c*d) / 2
-s3 = s1//s2
+s3 = s1 // s2
 ```
 
 **Spacing after commas?? add except in indexing?**
@@ -260,11 +260,11 @@ y = m*x + b  ! meters
 
 Use sentence case (as opposed to title case) for page and section titles.
 
-Use *emphasis* (`*emphasis*`/`_emphasis_`, rendered as italic) for key words/phrases, for emphasis, ...
+Use *emphasis* (`*emphasis*`/`_emphasis_`, rendered as italic) for key words/phrases when they are first introduced, for emphasis, ...
 
 Avoid use of **strong** (`**strong**`, rendered as bold) within paragraphs, since bold style is used for headings, drawing attention to examples (**Example:**), admonition/aside titles, etc.
 
 Make use of the admonition/aside [includes](_includes) (*note*, *tip*, *important*) where appropriate.
-* *note*: extra information
+* *note*: extra information, something that might appear in a footnote
 * *tip*: information about best practices, practical tips
-* *important*: warnings, etc.
+* *important*: warnings, things to avoid, etc.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -248,13 +248,27 @@ y4 = (a*b + c*d) / 2
 s3 = s1 // s2
 ```
 
-**Spacing after commas?? add except in indexing?**
+Generally add a space after commas, except when indexing with short index values or variables names.
+```fortran
+a(:,1)
+a2(1:10, 2:5)
+b(i,j)
+b2(long_i_name, long_j_name)
+b3(i + 2, j)
+call some_subroutine(a, b, an_option=.false.)
+c = [1, 2, 3, 10]
+d = [i, i = 1, 10]
+do i = 1, 10
+! ...
+```
 
 Capitalize the first letter for inline comments except for trailing inline comments that only consist of one word or a short phrase.
 ```fortran
 ! Compute new values
 y = m*x + b  ! meters
 ```
+
+These code style recommendations are similar to those in [the DFTB+ style guide](https://dftbplus-develguide.readthedocs.io/en/latest/fortranstyle.html).
 
 ### Text
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -92,8 +92,8 @@ This will force the GitHub content delivery network to serve you an updated vers
 
 ### Markdown
 
-- Use [code blocks](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code-and-syntax-highlighting),
-  denoted by back ticks (```), to surround code excerpts, programming language keywords, variables names and file names.
+- Place code excerpts in [code blocks](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code-and-syntax-highlighting),
+  denoted by back ticks (```` ``` ````). Use inline code style (`` `code` ``) for inline code excerpts, programming language keywords, variables names and file names.
 
 - Have no more than one sentence per source-code line, and break-up long sentences across multiples lines -
    this is important to avoid large git diffs and code review blocks on github.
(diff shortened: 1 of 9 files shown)


awvwgk:
I'm personally in the 3 space indentation fraction, but we usually consider using 2 or 4 spaces here. Most important is readability w.r.t. horizontal space IMO, the code block shouldn't get clamped on the right side due to over-indented code. The code blocks on the homepage can have a width of 80 to 90 characters before horizontal scrolling is necessary in our current page layout. This shouldn't be an issue in the quickstart guide but could be an issue for other minibooks which require more complex examples.

I think either two spaces and indentation of every unit (something, end something), like used now in the quickstart guide:

```f90
module m
  implicit none
  private
  public :: a_t

  type :: a_t
    integer :: val
  contains
    procedure :: func
  end type a_t

contains

  subroutine func(self)
    class(a_t), intent(in) :: self
    if (val > 0) then
      print *, val
    end if
  end function func

end module m
```

Or four space indentation only for non-procedure / module units (I think this is @certik's preferred style, I usually find this hard to read):

```f90
module m
implicit none
private
public :: a_t

type :: a_t
    integer :: val
contains
    procedure :: func
end type a_t

contains

subroutine func(self)
class(a_t), intent(in) :: self
if (val > 0) then
    print *, val
end if
end function func

end module m
```

zmoon:
> Most important is readability w.r.t. horizontal space IMO, the code block shouldn't get clamped on the right side due to over-indented code. The code blocks on the homepage can have a width of 80 to 90 characters before horizontal scrolling is necessary in our current page layout. This shouldn't be an issue in the quickstart guide but could be an issue for other minibooks which require more complex examples.

On mobile in portrait mode, the number of viewable columns can be even less. I am seeing that longer lines get soft-wrapped, which can be a bit confusing since the rendered blocks don't currently have line numbers. Perhaps optional code block line numbers or horizontal scroll bars could be added. I think 90 is a reasonable max and inline comments shouldn't go too far over it.

LKedward:
Many thanks for looking into this Zachary @zmoon . For most of your points, as long as the style is consistent across the tutorials then the particular choice doesn't matter too much. Re indentation, I think we should go with 2 spaces with zero base indentation for reasons already discussed, and my preference would be to indent module procedures for readability as suggested by Sebastian. I would also like to avoid vertical aligning of `::` and comments since this adds maintenance burden and pushes the line length in most cases. Once you settle on a set of rules for style, would you mind documenting them as a new section in the [Style Guide](https://github.com/fortran-lang/fortran-lang.org/blob/master/CONTRIBUTING.md#style-guide)? This will help others keep future edits consistent.



> * what is meant by "`do concurrent` is not a basic feature of Fortran" in _Operators and Control Flow_?

I'm not sure what is meant by this, please remove or reword as you see fit.



> * would a page on formatted printing and such fit here?

Yes absolutely, this would make a good addition to the quickstart tutorial.

milancurcic (approved):
Thank you, @zmoon, all looks good, pending the feedback from @awvwgk and @LKedward.

Even when we enable horizontal scroll, we should be conservative with indentation because of small line length on mobile (it's only 30- or 40-something characters on most smartphones).

zmoon:
Thanks @awvwgk, @LKedward, @milancurcic for the comments so far. I will draw up tutorial style guidelines for the [Style Guide](https://github.com/fortran-lang/fortran-lang.org/blob/master/CONTRIBUTING.md#style-guide) and then wait for feedback on that before enforcing it across the *Quickstart* pages.

I will do a separate PR for page on formatted print.

certik:
Yes for tutorials let's use zero based indentation and 2 spaces, essentially spare as much as we can, so that it's reasonable on mobile also.

CONTRIBUTING.md
```diff
@@ -185,3 +185,86 @@ populated by the `<h2>` headings on the current page.
...
+  contains
+    procedure :: func
+  end type a_t
+
+contains
+
+  subroutine func(self)
+    class(a_t), intent(in) :: self
+    if (val > 0) then
+      print *, val
```
awvwgk, line 220: Missed this in my example snippet.
```suggestion
    if (self%val > 0) then
      print *, self%val
```

zmoon:
Ok, I started the style guidelines ([here](https://github.com/zmoon/fortran-lang.org/blob/quickstart/CONTRIBUTING.md#tutorials)). Feel free to comment.

What about spaces after `,`? I don't think anyone commented on that yet. Currently there are many cases in the tutorial where no spaces are used after `,`, e.g. for `do` loops, array init, printing, etc. My personal preference is to add a space after `,` except in indexing (`a(:,:)`, `a(1,2)`, `do i = 1, n`, `b = [1, 2, 3]`). I feel that this usually improves readability.

LKedward:
Style guide additions look good to me, thanks! I think your suggestion for spacing around commas seems reasonable to me.
This PR is still in draft status - do let us know when you're ready for review. Great work!

zmoon:
@LKedward I still have a few things I want to do first but I will be sure to mark ready after that, thanks.

zmoon:
@milancurcic could you (or someone else) explain what is meant by "`do concurrent` is not a basic feature of Fortran" in Operators and Control Flow?

learn/quickstart/operators_control_flow.md
```diff
@@ -54,74 +54,76 @@ message to describe the nature of the `angle` variable:
...
+__Example:__ multi-branch  `if`-`else if`-`else`
 ```fortran
-  if (angle < 90.0) then
-    print *, 'Angle is acute'
-  else if (angle < 180.0) then
-    print *, 'Angle is obtuse'
-  else
-    print *, 'Angle is reflex'
-  end if
+if (age < 90.0) then
```
awvwgk, line 87:
```suggestion
if (angle < 90.0) then
```

awvwgk:
I think the comment on `do concurrent` refers the difference between expectation and actual implementations in compilers. I'll try to summarize what I picked up from the threads in discourse:
- concurrent doesn't mean parallel, yet most users mean parallel when they are using `do concurrent`
- from a strict viewpoint, there is no guarantee that a `do concurrent` can actually be run in parallel
  - the compiler must ensure that the loop is correctly run sequentially if no parallelization due to data dependencies is possible
  - practically this kind of compile time checks are difficult, if not impossible to implement (calling a `pure` procedure)
- some compiler implement the `do concurrent` as a parallel do loop regardless, even if there is a data dependency which could prevent parallelization
- some of those short-comings have been fixed in the Fortran 2018 variant of `do concurrent` which allows to declare the data dependencies explicitly

learn/quickstart/operators_control_flow.md
zmoon: Thanks for fixing that again!

zmoon:
> I think the comment on `do concurrent` refers the difference between expectation and actual implementations in compilers.

So maybe the first sentence in the note, "`do concurrent` is not a basic feature of Fortran.", could be replaced by something like "Don't expect a loop to be parallelized simply by changing from `do` to `do concurrent`". And I think also could mention compiler flags are needed, yes?

Is there a resource comparing what the different major compilers can/will do with it?

awvwgk:
Maybe we can just put a general statement there that concurrency does not imply parallel execution. Not sure what GFortran does, but `ifort` will parallelize `do concurrent` if the `-qopenmp` flag is set, `nvfortran` will parallelize with `-stdpar=multicore` and generate device offloading code with `-stdpar=gpu`.

zmoon:
Cool. Yeah I just feel since it's for the Quickstart we should mention that compiler flags are necessary to get parallelization, and maybe mention a few / link to an external resource.

zmoon:
@awvwgk I made some changes to the `do concurrent` note but feel free to suggest others

learn/quickstart/operators_control_flow.md
```diff
@@ -209,9 +209,11 @@ that any given loop iteration does not depend on the prior execution of other lo
...
 
 
-{% include important.html content="`do concurrent` is not a basic feature of Fortran. The explanation given does not detail
-all the requirements that need to be met in order to write a correct `do concurrent` loop. Compilers are also free to do as they see fit,
-which means they may not optimize the loop." %}
+{% include important.html content="Simply replacing a `do` loop with a `do concurrent` does not guarantee parallel execution.
+The explanation given does not detail all the requirements that need to be met in order to write a correct `do concurrent` loop.
+Compilers are also free to do as they see fit, meaning they may not optimize the loop (e.g., a small number of iterations doing a simple calculation).
+In general, compiler flags are required to activate possible parallelization for `do concurrent` loops.
+For `gfortran`, `-ftree-parallelize-loops=n` + an `-O` flag ≥ 1 activates auto-parallelization where `n` is the number of threads; for example, use `-ftree-parallelize-loops=0 -O3` to use all available threads." %}
```
awvwgk, line 216: It's easier and more readable to capture the content of an admonition
```suggestion
{% capture note %}

Simply replacing a `do` loop with a `do concurrent` does not guarantee parallel execution.
The explanation given does not detail all the requirements that need to be met in order to write a correct `do concurrent` loop.
Compilers are also free to do as they see fit, meaning they may not optimize the loop (e.g., a small number of iterations doing a simple calculation).
In general, compiler flags are required to activate possible parallelization for `do concurrent` loops.
For `gfortran`, `-ftree-parallelize-loops=n` + an `-O` flag ≥ 1 activates auto-parallelization where `n` is the number of threads; for example, use `-ftree-parallelize-loops=0 -O3` to use all available threads.

{% endcapture %}
{% include important.html content=note %}
```

zmoon: Ok, will do. How about the content though?

awvwgk: Is there any actual documentation for `-ftree-parallelize-loops` together with `do concurrent`? The closest I found so far is this bug report: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83064

zmoon: I couldn't find much besides the one you cite. And [this GCC Wiki page](https://gcc.gnu.org/wiki/AutoParInGCC), but it is brief and Fortran not mentioned.

awvwgk: Than I think we shouldn't advertise this option here.

zmoon: Ok, removed.

zmoon:
@LKedward do you want to review?

LKedward (approved):
Apologies for the delay in reviewing - this is great work @zmoon! A significant improvement and it's great to have a set guidelines for styling future mini-books. Very minor, but you could mention the use of the [Oxford comma](https://en.wikipedia.org/wiki/Serial_comma) in the _Text_ section of the style guide for future consistency. Other than that I haven't spotted any issues with these changes. Many thanks!

LKedward:
#build_preview

Gemfile.lock
```diff
@@ -58,4 +58,4 @@ DEPENDENCIES
   jekyll (~> 3.8.5)
 
 BUNDLED WITH
-   2.1.4
+   2.2.15
```
zmoon, line 61: This change was not intended, I will revert.

zmoon:
@LKedward were you able to look at those small CSS changes and their impact? It adds horizontal scroll bars for the code blocks e.g. for mobile. And I was trying to fiddle with the container width a bit, in regards to #256.

LKedward:
Sorry @zmoon, yes those CSS updates look fine to me and the rendering seems to work better now for narrow screens 👍

awvwgk (approved):

Gemfile.lock
awvwgk:
```suggestion
   2.1.4
```

zmoon: Thanks @awvwgk

awvwgk:
With three approvals I will go ahead and merge this patch.

awvwgk:
#delete_preview

PR #259, opened 2021-05-13
Adding packages

scivision:
Ondrej requested I add Gemini3D. I added / updated a few more packages.

### DIFF
--- a/_data/package_index.yml
+++ b/_data/package_index.yml
@@ -476,13 +476,13 @@
   version: none
 
 - name: h5fortran
-  github: scivision/h5fortran
+  github: geospace-code/h5fortran
   description: Simple, robust, thin HDF5 polymorphic read/write interface
   categories: io
   tags: hdf5
 
 - name: nc4fortran
-  github: scivision/nc4fortran
+  github: geospace-code/nc4fortran
   description: Object-oriented interface for NetCDF4 in Fortran
   categories: io
   tags: netcdf
@@ -1159,7 +1159,6 @@
   description: Easy examples of scientific computing with modern, powerful, easy Fortran 2018 standard
   categories: examples
   tags: block coarray contiguous mpi namelist openmp random submodule iso_fortran_env
-  license: GPL-2
 
 - name: Fortran patterns
   github: farhanjk/FortranPatterns
--- a/_data/package_index.yml
+++ b/_data/package_index.yml
@@ -803,6 +803,16 @@
   license: MIT
   version: none
 
+- name: MapTran3D
+  github: geospace-code/maptran3d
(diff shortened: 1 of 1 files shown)


_data/package_index.yml
```diff
@@ -628,6 +628,12 @@
   tags: blas linear-algebra
   license: BSD-3-Clause
 
+- name: MUMPS
+  github: scivision/mumps
+  description: parallel sparse direct solver, example of injecting CMake in non-CMake source
```
certik, line 633: I don't think I would put some build system specific things into the description, what do you think?
```suggestion
  description: parallel sparse direct solver
```

certik (commented):
I checked everything and it looks good, I think it passes our criteria.

However, there might be a license issue. The MUMPS package is under the CeCILL-C license and I don't see it here in the list of OSI approved licenses:

https://opensource.org/licenses/alphabetical

The only one I can see is CECILL-2.1, the full text is here:

https://opensource.org/licenses/CECILL-2.1

Compared to the CeCILL-C text:

https://cecill.info/licences/Licence_CeCILL-C_V1-en.html

there are some slight differences and I don't know if that matters or not.

I strongly think we should only allow OSI approved licenses. @milancurcic, @LKedward, @scivision and others, what do you think?

certik:
This precedent:

* https://github.com/cea-hpc/clustershell/commit/d8954586026510c884e0ee4240dbffc67444d35b
* https://github.com/cea-hpc/clustershell/issues/321

seems to suggest that the CeCILL-C license might not be OSI approved...

certik:
Ok, so according to this:

* https://spdx.org/licenses/

The CECILL-C license is not OSI approved. 

Here is what I suggest we do:

* Remove MUMPS from this PR, then there is no problem with this PR, let's get it merged
* Create a new PR for MUMPS to discuss what to do next

I suggest to only allow OSI (https://opensource.org/licenses/alphabetical) approved licenses to be allowed in the main section.

Should we have another section that allows non OSI approved software written in Fortran? In this section we could list all kinds of commercial codes also, such as VASP. We could put MUMPS there.

certik (approved):
This looks good to me to merge as is.

scivision:
Yes I think it will be necessary to have a second list for non-OSI software. Given a few factors such as the long lifetime of Fortran, and the nature of the aerospace and other such government and research organizations that make Fortran code. To protect (perhaps misguidedly) their IP and commercial relicensing ability they have what are open source but not necessarily free licenses. Some recently released software has some of the most restrictive licenses I've seen on open source packages. But the software is essential for a range of geospace research.

certik:
Thanks!

Yes, there will be quite a few packages that we could list there. I created #260 to discuss that.

PR #273, opened 2021-06-04
add the SNaC package ?

p-costa:
Hello everyone,

I wrote [SNaC](https://github.com/p-costa/SNaC), a multi-block finite-difference solver for massively parallel simulations of fluid flows. Despite being young and the low number of stars, I hope you find it interesting and suitable for integrating the list of scientific codes.

Thanks in advance!
Pedro

### DIFF
--- a/_data/package_index.yml
+++ b/_data/package_index.yml
@@ -1166,7 +1166,13 @@
   github: gemini3d/gemini3d/
   description: 3D ionospheric model covering local (> 10 gyroradii) to global scales.
   categories: scientific
-
+  
+- name: SNaC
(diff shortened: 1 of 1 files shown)


certik:
Thanks @p-costa for submitting it! Is the "the primary functionality of the package" implemented? If so, then I think it should be included as far as I am concerned.

p-costa:
Thanks, Ondřej @certik! Yes, it is definitely mature regarding the funtionality.

certik (approved):
+1 from me.

milancurcic (approved):
Looks great, thanks!

PR #286, opened 2021-07-02
Installation process for GFortran on OpenBSD

brocolis:
Installation process for GFortran on OpenBSD.

### DIFF
--- a/learn/os_setup/install_gfortran.md
+++ b/learn/os_setup/install_gfortran.md
@@ -98,7 +98,12 @@ sudo port install gcc10
 
 ## OpenBSD
 ```bash
-pkg_add g95-8.4.0p6
+pkg_add g95
+```
+
+On OpenBSD, the GFortran executable is named `egfortran`. To test it, type:
+```bash
+egfortran -v
 ```
 
 ---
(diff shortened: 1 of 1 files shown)


LKedward (approved):
Thanks Carlos @brocolis - I don't have any experience with OpenBSD, but these changes look good to me.

brocolis:
Thanks for reviewing it, @LKedward. I have one day of experience with OpenBSD 😸. Today I discovered that we can install GFortran with `pkg_add g95`, no need to include full version number. I'll update the PR.

learn/os_setup/install_gfortran.md
```diff
@@ -95,6 +95,17 @@ Install a gcc version:
...
 ```
+
+## OpenBSD
+```bash
+pkg_add g95
+```
+
+On OpenBSD, the GFortran executable is named `egfortran`. To test it, type:
+```bash
+egfortran -v
```
awvwgk, line 106: Is this an issue for fpm to discover the right compiler? It probably requires to set `--compiler egfortran` on every as well.

brocolis: @awvwgk, yes, `fpm` works with `--compiler egfortran`.  `export FPM_COMPILER=egfortran` also works 👍.

awvwgk (approved):

milancurcic:
Very nice, thanks @brocolis!

PR #290, opened 2021-07-10
Add arrayfire-fortran to package index

awvwgk:
Repository: https://github.com/arrayfire/arrayfire-fortran

See also https://fortran-lang.discourse.group/t/questions-from-a-fortran-hpc-webinar/1516/20

### DIFF
--- a/_data/package_index.yml
+++ b/_data/package_index.yml
@@ -209,7 +209,13 @@
   github: swig-fortran/flibcpp
   description: Fortran bindings to the C++ Standard Library
   categories: interfaces
-  tags: c++
+  tags: cpp
+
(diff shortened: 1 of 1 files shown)


LKedward (approved):
Thanks, I didn't know this existed! I will have to try this out at some point. Though the latest commit is a bit old now.

awvwgk:
Looks like arrayfire has binding to almost every language (I see Haskell, Ruby, R, Lisp, ...). With that many bindings it is probably difficult to keep them all up to date, maybe if we express interest in the Fortran binding they will consider updating it again.

certik (approved):

milancurcic:
Thank you!

PR #294, opened 2021-07-21
compilers: use more objective tone

scivision:
The current verbiage for a couple compliers sounds like an advertisement vs. a reference article

### DIFF
--- a/compilers.md
+++ b/compilers.md
@@ -29,7 +29,7 @@ This project is under active development.
 
 ### Current Flang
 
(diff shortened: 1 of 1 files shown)


certik:
Yeah --- we took it from their advertising materials. Ultimately I want this to be left to them to provide a text under 100 (?) words to put on the webpage. How to ensure it is less advertising and more information? I don't know. Until then, we can modify it along the lines of this PR. I think that will help show what we are looking for.

certik (approved):

milancurcic (approved):
Looks good, thanks!

PR #296, opened 2021-07-26
my software with at least 5 stars

MarDiehl:
I still plan to put os into stdlib, but currently don't have time.


certik (approved):
I plan to do the same with my code, but also currently don't have time. But eventually we will make it happen. :)

awvwgk (approved):

milancurcic (approved):
Looks good, thanks!

PR #297, opened 2021-07-27
Fix insecure workflow.

intrigus-lgtm:
Actions triggered on "issue_comment" have access to
write permissions and repository secrets by default.
If we checkout and build untrusted code an attacker
can get access to the secrets as well. They can exe-
cute code by adding a jekyll ruby plugin in
"_plugins/malicious.rb".

To prevent this, we split the building of the site
and pushing the site to gh-pages and commenting on
the PR.

The site building is done with only read
permissions, so running untrusted code is not a
problem. If the site building finished successfully
a privileged action is triggered that then pushed
the site to the gh-pages branch and comments on the
PR.

I also added a check that prevents pushing 
".github/workflows/" files to the gh-pages branch, because
this would allow secret extraction and repository modification as well.
(If you can add new workflows files you do what every you want inside them)

### DIFF
--- a/.github/workflows/buildPRPreview.yml
+++ b/.github/workflows/buildPRPreview.yml
@@ -1,8 +1,7 @@
 # Github action to build Jekyll site for a pull request preview
-#
-#  Built site is pushed to 'gh-pages' branch
-#   under directory pr/pr_id/
-#
+# We have to checkout and run untrusted code! This is dangerous if done in a privileged context.
+# That's why this action only has read-permissions.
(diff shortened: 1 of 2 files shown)


certik:
@intrigus-lgtm thank you so much for this. We all really appreciate it! I'll let @LKedward review it.

LKedward (approved):
Thanks again for this. As per our previous conversation, this all looks good to me 👍

awvwgk (approved):

milancurcic (approved):
Thank you @intrigus-lgtm!

PR #320, opened 2021-09-07
Alternative approach to avoiding integer division

gareth-nx:
The tutorial suggests bypassing integer division by type-casting one variable to real, i.e. to avoid `3/4 == 0`, you can do `real(3, dp) / 4`. 

While that's fine, it is often nicer to simply write `3 * 1.0_dp / 4`. This pull request updates the tutorial to note that.

In addition, the documentation now notes Fortran's default left-to-right operator precedence -- which affects one of the examples on the page about integers.

### DIFF
--- a/learn/best_practices/floating_point.md
+++ b/learn/best_practices/floating_point.md
@@ -74,12 +74,19 @@ a = 3
 ```
 
 In order to impose floating point division (as opposed to integer
-division `1/2` equal to `0`), one can convert the integer to a floating
+division `3/4` equal to `0`), one can convert the integer to a floating
 point number by:
 
 ```fortran
 real(dp) :: a
-a = real(1, dp) / 2  ! 'a' is equal to 0.5_dp
+a = real(3, dp) / 4  ! 'a' is equal to 0.75_dp
+```
+
+or simply separate the integer division with multiplication by `1.0_dp`
+
+```fortran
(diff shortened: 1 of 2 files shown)


brocolis (approved):

awvwgk (approved):

Beliavsky:
Why would you write `3 * 1.0_dp / 4` instead of just `3.0_dp/4` ?

gareth-nx:
@Beliavsky Indeed you would not! 

However, if the `3` and `4` were actually variables (say `i` and `j`), then the approach presented in the edits generalises correctly `(i * 1.0_dp / j)`. 

Your comment also applies to the other example -- we wouldn't write `real(3, dp) / 4` in preference to `3.0_dp / 4`. But the former version generalises.

That was my thinking. But it's tricky to have just the right amount of complexity without confusing people. Suggestions for making it clearer are most welcome.

gareth-nx:
One approach -- we could change the example to define `i = 3; j = 4`, so that the use of variables is clear. But I don't know whether the extra complexity is worth the benefit.

gareth-nx:
I looked at this again, and could not see a clean way to introduce integer variables `i, j` without adding too much complexity (noting the style of the tutorial).  

So in my opinion this is ready to merge.

PR #335, opened 2021-09-24
Adding two more packages to the index

tclune:


_data/package_index.yml
```diff
@@ -1375,3 +1375,17 @@
...
   tags: game
+
+- name: gFTL
+  github: Goddard-Fortran-Ecosystem/gFTL
+  description: Templates for  STL-like Fortran implementations of containers for arbitrary types (Vector, Set, Map, Deque, Stack, Queue, ...)
+  categories: containers 
+  tags: container template
+  license: Apache-2.0 
+  
+  - name: gFTL
```
milancurcic, line 1386: The `name` field seems incorrect. Perhaps a bad copy pasta? :spaghetti:

tclune:
```suggestion
  - name: yaFyaml
```
Yes - sorry about that.   Could have sworn I had updated that field.   Probably should not have done this while listening to the talk.  :-)

milancurcic (approved):
Thanks @tclune!

_data/package_index.yml
```diff
@@ -1375,3 +1375,17 @@
   description: Object-oriented Fortran falling block game using Curses console graphics
   categories: examples
   tags: game
+
+- name: gFTL
+  github: Goddard-Fortran-Ecosystem/gFTL
+  description: Templates for  STL-like Fortran implementations of containers for arbitrary types (Vector, Set, Map, Deque, Stack, Queue, ...)
```
certik, line 1381:
```suggestion
  description: Templates for STL-like Fortran implementations of containers for arbitrary types (Vector, Set, Map, Deque, Stack, Queue, ...)
```

certik (approved):
Very cool, thanks @tclune !

tclune:
I just remembered that I should add in the gFTL-shared project which instantiates the templates for common containers with intrinsic elements.     As a project it is uninteresting, but for users, it might be all they need.  Kept separate because the templates don't require any specific compiler and can be installed in a compile-independent location.

PR #351, opened 2021-12-12
Use HEAD to reference default branch

awvwgk:
- reference HEAD commit instead of branch name


milancurcic:
I just noticed this--thank you!

mwtoews:
This is now causing 404: Not Found errors. I believe it was correct before.

PR #357, opened 2021-12-31
January 2022 newsletter draft

milancurcic:
Sorry for the last-minute submission. Considering people may be offline around New Year's eve, let's plan to complete and merge this by Wednesday, January 5, if possible

* [x] PRs summary
* [x] GitHub stats data
* [x] Flang updates by @AlexisPerry 
* [x] LFortran updates by @certik

PR #358, opened 2022-01-01
Fix end time calculation in the PRs script

milancurcic:
While preparing #357 I found a small issue in the PRs script which manifests only in December. This PR fixes it and adds a requirements.txt file to keep track of the script dependencies.

We should also have a short guide on making a newsletter, which is for another PR.


milancurcic:
I messed up my branch, it shouldn't include these extra commits. Hold on.

PR #359, opened 2022-01-01
Fix time calculation in the PRs script

milancurcic:
While preparing #357 I found a small issue in the PRs script which manifests only in December. This PR fixes it and adds a requirements.txt file to keep track of the script dependencies.

We should also have a short guide on making a newsletter, which is for another PR.


zmoon (commented):
Oops, sorry about that, and thanks for fixing! We could do it with div-rem math and avoid `dateutil`, but that would be less clear.

_scripts/prs.py
```diff
@@ -3,6 +3,7 @@
 Requires PyGithub.
```
zmoon, line 3: could add `python-dateutil` to this note? or remove the line, as maybe it won't be necessary after the documentation you mentioned is added

_scripts/prs.py
```diff
@@ -3,6 +3,7 @@
...
 """
 import datetime
+from dateutil.relativedelta import relativedelta
```
zmoon, line 6: convention would be to put this after `import github`, to stress that it is not part of the std lib

awvwgk: I recommend to just run a formatter (`black` or similar) over the script.

_scripts/prs.py
milancurcic: Yes. I've added requirements.txt which tracks dependencies, but we haven't yet documented this. So for now, I think it's good to mention it here, until it's documented elsewhere.

_scripts/prs.py
milancurcic: Either sounds good. I haven't used formatters before.

zmoon: could add a pre-commit config to do it automatically

milancurcic: I ran it through black. Let me know if there's anything else.

zmoon:
Maybe the `requirements.txt` should go inside the `_scripts` directory?

milancurcic:
OK, done, let me know if it's good to go.

zmoon (approved):

Issue #217, opened 2021-03-03
Search for c++ packages does not work correctly


jme52:
One of the "Featured topics" on https://fortran-lang.org/packages/ is c++.  However, the search page https://fortran-lang.org/packages/search/?query=c++ seems to list many packages that don't have c++ in their list of tags or descriptions.  Is it possible that the search inputs are not properly escaped?

henilp105:
The Featured topics has changed it to cpp as that '+' in c++ was being termed as a space replacer and to join a line of words
by the script thus only rendering the results for "c  " that is c with 2 spaces. that has been fixed by replacing c++ by cpp.
request to close this issue as it has been resolved.

Thanks and Regards,
Henil Panchal

henilp105:
@awvwgk  @jme52  This issue has been resolved thus request to close this issue #217.

Thanks and Regards,
Henil Panchal

awvwgk:
I think this issue is still present, since we do not properly escape a search like `c++` at the moment.

henilp105:
sir I can make the search code so should i contribute it.
Thanks and Regards,
Henil Panchal

awvwgk:
The package search is implemented here: https://github.com/fortran-lang/fortran-lang.org/blob/master/assets/js/package_search.js

PR #364, opened 2022-01-18
Resolves Issue #217

henilp105:
I have made a function that does the Regex and Resolves the issue for c++.

Thanks and Regards,
Henil Panchal


henilp105:
#build_preview

PR #366, opened 2022-01-18
Resolves Issue #217 search packages

henilp105:
I have made a function that would handle all the search utilities and regex and this also solves the Issue #217.

Thanks and Regards,
Henil Panchal


henilp105:
#build_preview

henilp105:
#delete_preview

PR #371, opened 2022-01-29
Add link to "The 'F' Word" video series

awvwgk:
See https://fortran-lang.discourse.group/t/2679

cc @fluidnumerics-joe

### DIFF
--- a/_data/learning.yml
+++ b/_data/learning.yml
@@ -119,7 +119,7 @@ reference-courses:
   - name: "PRACE Course: Advanced Fortran Topics"
     url: https://doku.lrz.de/display/PUBLIC/PRACE+Course%3A+Advanced+Fortran+Topics
 
-  - name: "The 'F' Word"
+  - name: "The 'F' Word - Programming in Fortran"
     url: https://www.youtube.com/playlist?list=PLRO4xf5MdhAv9CNTETor75rANZtBqPVgQ
     description: >-
       modern Fortran for developing an extensible library that can be used to solve conservation laws (PDEs) using spectral and spectral element methods
--- a/_data/learning.yml
+++ b/_data/learning.yml
@@ -119,6 +119,11 @@ reference-courses:
   - name: "PRACE Course: Advanced Fortran Topics"
     url: https://doku.lrz.de/display/PUBLIC/PRACE+Course%3A+Advanced+Fortran+Topics
 
+  - name: "The 'F' Word"
+    url: https://www.youtube.com/playlist?list=PLRO4xf5MdhAv9CNTETor75rANZtBqPVgQ
+    description: >-
+      modern Fortran for developing an extensible library that can be used to solve conservation laws (PDEs) using spectral and spectral element methods
+
   - name: "Kursmaterial für Wissenschaftliches Programmieren (Modern Fortran, 2017)"
     url: https://www.bccms.uni-bremen.de/cms/people/b-aradi/wissen-progr/modern-fortran/2017
     description: >-


fluidnumerics-joe:
@awvwgk , Thanks for making this suggestion!

milancurcic (approved):
Thanks @awvwgk, and of course thanks @fluidnumerics-joe for making these great video series!

certik:
@fluidnumerics-joe is there any alternative name that we could use? How about "The 'Fortran' Word"?

I personally feel this title is unfortunate as it coincides with a word that we should not be using here. If others think this title is ok, then I am fine to be outvoted on this.

awvwgk:
> I personally feel this title is unfortunate as it coincides with a word that we should not be using here. If others think this title is ok, then I am fine to be outvoted on this.

I think this coincidence is wished. In the context I think it is quite clear what is meant. However, we can certainly adjust the description to make it clearer. Suggestions are welcome, I just copied a bit from Joe's post on discourse.

milancurcic:
I had the same thought for a few seconds but I think it's fine. I find it cheeky and smart.

_data/learning.yml
```diff
@@ -119,6 +119,11 @@ reference-courses:
   - name: "PRACE Course: Advanced Fortran Topics"
     url: https://doku.lrz.de/display/PUBLIC/PRACE+Course%3A+Advanced+Fortran+Topics
 
+  - name: "The 'F' Word"
```
certik, line 122: Let's use the wording from here:

https://twitter.com/fortranlang/status/1487504884865904644

```suggestion
  - name: "The 'F' Word - Programming in Fortran"
```

certik:
I bet every reader of this will have the same thought. I provided a suggestion above. I just wanted to check on this, but it looks like it's fine with others, so I am ok overall.

_data/learning.yml
fluidnumerics-joe: @certik -  I think your suggestion is a good one. My intention is to reflect the observation that computer scientists are often offended (in some way) about developing new applications in Fortran these days. Adding the clarifier "Programming in Fortran" definitely helps affirm which offensive 'F' word I mean.

certik (approved):
Awesome. This is fine with me now.

Yes, Fortran is in an interesting position, as there is so much bias against it. And then once people actually try it (with an open mind), they realize it's actually a pretty good language!

fluidnumerics-joe:
@certik, much agreed. Thanks for the suggestion to keep this a bit more palatable.

certik:
@fluidnumerics-joe thanks for all your videos! Fortran needs that.

Once LFortran gets to beta (from alpha), this year, I am hoping to collaborate with you and others who have Fortran codes. There is all kinds of cool work that we could do, from GPU support to optimizations. We are adding some excellent optimizations, we just merged this one: https://gitlab.com/lfortran/lfortran/-/merge_requests/1649, which optimizes `if (modulo(int(Nd), 2) == 1) x = -x` into floating point "bit hacks" in LLVM/assembly using shift and xor. GFortran wasn't able to do that. Once we have enough of such optimizations, I think people will be very excited with the performance.

ivan-pi:
I also found the title smart and provocative in the given context.

awvwgk:
#build_preview

awvwgk:
Thanks everybody, I'm going ahead and merge.

#delete_preview

fluidnumerics-joe:
> Once LFortran gets to beta (from alpha), this year, I am hoping to collaborate with you and others who have Fortran codes. There is all kinds of cool work that we could do, from GPU support to optimizations. We are adding some excellent optimizations, we just merged this one: https://gitlab.com/lfortran/lfortran/-/merge_requests/1649, which optimizes `if (modulo(int(Nd), 2) == 1) x = -x` into floating point "bit hacks" in LLVM/assembly using shift and xor, which we needed for a very fast `sin(x)` function implementation. GFortran wasn't able to do that. Once we have enough of such optimizations, I think people will be very excited with the performance.

@certik I'd love to dive into LFortran with you all. I recall some time ago, we had briefly discussed Focal (@LKedward ) and the possibility of supporting device subroutines and other Fortran native approaches for GPU acceleration. That being said, I'm still game for the Fortran-C interoperability approach.

PR #383, opened 2022-03-18
Updated VS Code extensions

gnikit:
Modern Fortran now supports natively Language Server support (FORTRAN Intellisense), formatting (fprettify and findent)

### DIFF
--- a/learn/os_setup/text_editors.md
+++ b/learn/os_setup/text_editors.md
@@ -25,18 +25,11 @@ Things to consider before choosing a text editor:
 ## Configuring VS Code
 
 With the editor opened, at the main UI, at the buttoned-column on the left, there is a *four-square-shaped* icon to open the Marketplace for extensions. 
-The following extensions are suggested for Fortran developement:
-- [Modern Fortran by Miguel Carvajal](https://marketplace.visualstudio.com/items?itemName=krvajalm.linter-gfortran)
-- [FORTRAN IntelliSense by Chris Hansen](https://marketplace.visualstudio.com/items?itemName=hansec.fortran-ls)
-    - Depends on the previous *Modern Fortran* extension, as well as [Python](https://www.python.org/) and [Fortran Language Server](https://github.com/hansec/fortran-language-server), which need to be installed.
-- [Fortran Breakpoint Support by ekibun](https://marketplace.visualstudio.com/items?itemName=ekibun.fortranbreaker)
-
-*Note: Both Modern Fortran and Fortran Breakpoint Support require ms-vscode.cpptools, for that you will also need to install the [C/C++ by Microsoft](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools) extension.*
+Install [Modern Fortran by Miguel Carvajal](https://marketplace.visualstudio.com/items?itemName=krvajalm.linter-gfortran) for syntax highlighting, linting and Language Server support.
 
 The following are not essential but more advanced users might find useful:
 - [Remote - WSL by Microsoft](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl)
 - [GitLens - Git supercharged by Eric Amodio](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens)
-- [fprettify by Blamsoft](https://marketplace.visualstudio.com/items?itemName=Blamsoft.fprettify)
 - [Even Better TOML by tamasfe](https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml). For Fortran developers that are using the new [fpm](https://github.com/fortran-lang/fpm) *Fortran Package Manager*, a TOML language support might come in handy.
 
 ## Configuring SublimeText
--- a/learn/os_setup/text_editors.md
+++ b/learn/os_setup/text_editors.md
@@ -25,11 +25,15 @@ Things to consider before choosing a text editor:
 ## Configuring VS Code
 
 With the editor opened, at the main UI, at the buttoned-column on the left, there is a *four-square-shaped* icon to open the Marketplace for extensions. 
-Install [Modern Fortran by Miguel Carvajal](https://marketplace.visualstudio.com/items?itemName=krvajalm.linter-gfortran) for syntax highlighting, linting and Language Server support.
+Install [Modern Fortran](https://marketplace.visualstudio.com/items?itemName=krvajalm.linter-gfortran) for syntax highlighting, linting and Language Server support and formatting.
 
-The following are not essential but more advanced users might find useful:
-- [Remote - WSL by Microsoft](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl)
-- [GitLens - Git supercharged by Eric Amodio](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens)
+### Additional Tools
+
+The following extensions are not essential but most users might find them useful:
+
+- [GitLens](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens) for advanced Git visualization and operations.
+- [CMake](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cmake-tools) for full-featured CMake integration in VS Code.
+- [Remote Development](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.vscode-remote-extensionpack) for working with containers, on a remote machines, or in the Windows Subsystem for Linux (WSL).
 - [Even Better TOML by tamasfe](https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml). For Fortran developers that are using the new [fpm](https://github.com/fortran-lang/fpm) *Fortran Package Manager*, a TOML language support might come in handy.
 
 ## Configuring SublimeText


certik:
Should we still keep the other links in a section like "Other tools", so that people can discover them?

gnikit:
I think the "Other tools" section should be kept mostly because VSCode without GitLens for big projects is rough.
The  extensions that I would vouch for are:

- GitLens which is crucial IMO
- [Remote Development](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.vscode-remote-extensionpack) for SSHing into workstations, WSL and spinning up containers

I would consider adding the [CMake](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cmake-tools) extension from Microsoft simply because a lot of people might find it useful (I prefer the command line, but to each their own).

I am not familiar with the TOML extension but I suspect it might be useful for fpm dev work

certik:
If you want, just put the links to the "Other Tools" or "Additional Tools" section, and we can merge it.

gnikit:
So I think this is a good place for new VS Code users that want to code in Fortran to start. I believe it should be okay now.

learn/os_setup/text_editors.md
```diff
@@ -25,18 +25,15 @@ Things to consider before choosing a text editor:
...
-- [fprettify by Blamsoft](https://marketplace.visualstudio.com/items?itemName=Blamsoft.fprettify)
+Install [Modern Fortran](https://marketplace.visualstudio.com/items?itemName=krvajalm.linter-gfortran) for syntax highlighting, linting and Language Server support and formatting.
+
+### Additional Tools
+
+The following extensions are not essential but most users might find them useful:
+
+- [GitLens](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens) for advanced Git visualization and operations.
+- [CMake](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cmake-tools) for full-featured CMake integration in VS Code.
+- [Remote Development](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.vscode-remote-extensionpack) for working with containers, on a remote machines, or in the Windows Subsystem for Linux (WSL).
```
certik, line 36: Let's list all the links for now:
```suggestion
- [Remote Development](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.vscode-remote-extensionpack) for working with containers, on a remote machines, or in the Windows Subsystem for Linux (WSL).
- [FORTRAN IntelliSense by Chris Hansen](https://marketplace.visualstudio.com/items?itemName=hansec.fortran-ls)
    - Depends on the previous *Modern Fortran* extension, as well as [Python](https://www.python.org/) and [Fortran Language Server](https://github.com/hansec/fortran-language-server), which need to be installed.
- [Fortran Breakpoint Support by ekibun](https://marketplace.visualstudio.com/items?itemName=ekibun.fortranbreaker)
- [Remote - WSL by Microsoft](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl)
- [fprettify by Blamsoft](https://marketplace.visualstudio.com/items?itemName=Blamsoft.fprettify)
```
I don't know if they are really obsolete, so let's just list them for now.

gnikit: thanks for the comments, pls see my updated comment about each individual extension

certik: I see. How about this then:

```suggestion
- [Remote Development](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.vscode-remote-extensionpack) for working with containers, on a remote machines, or in the Windows Subsystem for Linux (WSL).

The following extensions might be incompatible with the above Fortran extension, so you would most likely need to uninstall it to install these:

- [FORTRAN IntelliSense by Chris Hansen](https://marketplace.visualstudio.com/items?itemName=hansec.fortran-ls)
    - Depends on the previous *Modern Fortran* extension, as well as [Python](https://www.python.org/) and [Fortran Language Server](https://github.com/hansec/fortran-language-server), which need to be installed.
- [Fortran Breakpoint Support by ekibun](https://marketplace.visualstudio.com/items?itemName=ekibun.fortranbreaker)
- [Remote - WSL by Microsoft](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl)
- [fprettify by Blamsoft](https://marketplace.visualstudio.com/items?itemName=Blamsoft.fprettify)
```

gnikit: It is not exactly that they are incompatible (some of them are, FORTRAN Intellisense for sure & probably fprettify) but more like they were made as standalone extensions for features that are normally part of the main language extension, so if you have Modern Fortran installed they are all redundant

In more detail for each individual extension:

- "FORTRAN Intellisense" cannot be a standalone extension and it needs a parent extension e.g. Modern Fortran, which will associate the Fortran source code as files to be parsed and served by the, now dead, `fortran-language-server`. Effectively, this extension is only an interface for the Language Server Protocol and VS Code and was created, I think, because we (Modern Fortran) was doing it poorly back in 2017 or so. The "FORTRAN Intellisense" extension has outdated VS Code args and will try and install the old Language Server, which can become very messy for `fortls` users.
- "Remote - WSL" is actually included in the extension pack that I added called "Remote Development", which should be installing 3 extensions 1. for SSH 2. for Containers 3. for WSL
- "Fortran Breakpoint Support" is not a real VS Code extension i.e. it does not execute any code, but it takes the C++ debuggers only GDB and MSVC, has not been updated for LLDB, and relabels them as Fortran. I remember this used to cause some problems with C/C++ projects where the debuggers would display with the wrong name i.e. registered as Fortran. In addition, as far as I can remember, the way they interface with the VS Code API has been deprecated for many years now (+2-3y). I tried testing the extension now but I was unable to get it working in my VS Code Insiders build. The modern and "correct" way of registering debugger support for VS Code extensions I added in Modern Fortran in 2019 (I think, it's been a while since I did that).
- "fprettify" does not have a public repo, but what I suspect it does is [registering a formatting provider](https://code.visualstudio.com/blogs/2016/11/15/formatters-best-practices) which is what all extensions that provide formatting have to do. Having had a quick look, there are a couple more standalone formatting extensions in VS Code both going by the name `vscode-modern-fortran-formatter`. What these extensions do in the background is spawn a shell and launch `fprettify`, the same thing albeit fully-async is done in Modern Fortran with the formatters: `fprettify` and `findent` so the extension "fprettify"

For extensions that register things; "fprettify" a `DocumentFormattingEditProvider` and "FORTRAN Intellisense" a `LanguageClient` we have to also worry about these multiple registrations interfering with the main Modern Fortran extension. Theoretically this is allowed by the VS Code API but what happens in practice when no precedence is specified amongst multiple providers is *probably* unspecified behaviour.

There is one more major problem which I failed to mention before, the language ID used to identify Fortran source code is provided by the main Fortran extension i.e. Modern Fortran or fortran. The peripheral extensions discussed here, have these IDs hardcoded into them in order for them to work. In Modern Fortran I actually changed the Fixed Fortran ID from a weird name to one matching the Free Form ID, this effectively means that all these tools will not work for Fixed Form Fortran Code since they haven't been updated in years.

Now that is if people use Modern Fortran as their base extension. As I mentioned before, there is another extension called [`fortran`](https://marketplace.visualstudio.com/items?itemName=Gimly81.fortran) last updated in 2017, which can also provide its own Fortran Language ID, I think it also provides some basic syntax highlighting and that is it. If someone is using that, then they would need "fprettify" for formatting and "Fortran Breaking Support" for breaking points (which would also introduce all the issues I mentioned). The Language Server support would be limited to the old server (fortran-language-server) since "FORTRAN Intellisense" does not interface nicely with my version of the Language Server, simply because I have added new options that are not reflected in the "FORTRAN Intellisense" VS Code options.

However, because all these tools (fortran, FORTRAN Intellisense, frettify and Breaking points) have not been updated in years, I would recommend all newcomers to avoid them at all costs. I also understand that I am **very biased**, since I develop/maintain both Modern Fortran and fortls.

## TLDR

All features that these extensions provided are also present, up to date, and tested, in the Modern Fortran extension thus making them redundant. 

## Relying on multiple extensions for language features

In general and IMO, it does not make sense to create extensions that introduce a single feature like formatting, hovering, snippets, atocomplete etc.. since they are unable to function independently and always require other base extensions to provide the entire VS Code interface minus that one feature.

-----

I hope this clears things up a bit. I tried to condense 4+ years of VS Code Fortran dev-work into a single comment. Let me know if you think about what should be included in the web-page

certik: I have not used the extensions, so I don't know their status. I don't have a strong feeling about this, I thought mentioning them would not hurt, but it seems you don't want them even mentioned. I guess that's fine too, as long as there are no objections from others.

Overall I agree we should focus on ideally just one extension that just works for Fortran, and document how to get everything working.

certik: CC @milancurcic, @LKedward, @awvwgk.

awvwgk: I understand and share Ondřej's objection to delete references to other potentially alternative projects, especially if done by one of the upstream developers.

I'm not a VS Code user nor an extension developer, so I can hardly judge how those projects will interact with each other and whether it is useful or not to have them around. After a look to the version history of all extensions currently on the webpage, only Modern Fortran (`krvajalm.linter-gfortran`) is currently active, the others stopped being updated in 2020 and  some of them have a backlog of open issues and patches which are not handled at the moment. I guess this qualifies as dead projects as far as I'm concerned.

I'm fine with updating the page to only link to Modern Fortran, it seems like the best choice for VS Code users at the moment.

awvwgk (approved):

LKedward (approved):

awvwgk:
Thanks for the discussion and feedback everybody. I think we can go ahead and merge.

PR #389, opened 2022-04-14
Add librsb to package index

ivan-pi:
From the [website](http://librsb.sourceforge.net):

> librsb is a library for sparse matrix computations featuring the Recursive Sparse Blocks (RSB) matrix format. This format allows cache efficient and multi-threaded (that is, shared memory parallel) operations on large sparse matrices. The most common operations necessary to iterative solvers are available, e.g.: matrix-vector multiplication, triangular solution, rows/columns scaling, diagonal extraction / setting, blocks extraction, norm computation, formats conversion. The RSB format is especially well suited for symmetric and transposed multiplication variants. Most numerical kernels code is auto generated, and the supported numerical types can be chosen by the user at build time. librsb can also be built serially (without OpenMP parallelism), if required.
>
> librsb also implements the Sparse BLAS standard, as specified in the [[BLAS Technical Forum](http://www.netlib.org/blas/blast-forum/)] documents.

The package fulfills the required criteria:

* Relevance: complete implementation of the Sparse BLAS interface in Fortran
* Maturity: extensive testing has been done over several years
* Availability: http://librsb.sourceforge.net
* Open source: LGPL-3.0 or later
* Uniqueness: rare package for sparse matrix computations with a special storage format and one of less than a handful of Sparse BLAS implementation

cc @michelemartone, would you like to add any tags? I don't think any other [categories](https://github.com/fortran-lang/fortran-lang.org/blob/master/PACKAGES.md#github-hosted-packages) apply.

### DIFF
--- a/_data/package_index.yml
+++ b/_data/package_index.yml
@@ -254,6 +254,13 @@
   categories: interfaces
   tags: cpp gpu opencl
 
+- name: librsb
+  url: http://librsb.sourceforge.net/
+  description: A shared memory parallel sparse matrix computations library for the Recursive Sparse Blocks format 
+  categories: numerical interfaces scientific
+  tags: openmp sparse-matrix sparse-blas linear-algebra
+  license: LGPL-3.0-or-later
+  
 # --- Compiler
 
 - name: hipfort
--- a/_data/package_index.yml
+++ b/_data/package_index.yml
@@ -256,7 +256,7 @@
 
 - name: librsb
   url: http://librsb.sourceforge.net/
-  description: A shared memory parallel sparse matrix computations library for the Recursive Sparse Blocks format implementing the Sparse BLAS
+  description: A shared memory parallel sparse matrix computations library for the Recursive Sparse Blocks format implementing the Sparse BLAS standard
   categories: numerical interfaces scientific
   tags: linear-algebra openmp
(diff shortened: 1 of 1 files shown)


michelemartone:
It seems all correct to me.
Maybe as description you could better append " implementing the Sparse BLAS" so to have "A shared memory parallel sparse matrix computations library for the Recursive Sparse Blocks format implementing the Sparse BLAS" which is a bit more informative.
Thanks @ivan-pi !

_data/package_index.yml
```diff
@@ -254,6 +254,14 @@
   categories: interfaces
   tags: cpp gpu opencl
 
+- name: librsb
+  url: http://librsb.sourceforge.net/
+  description: A shared memory parallel sparse matrix computations library for the Recursive Sparse Blocks format implementing the Sparse BLAS
```
ivan-pi, line 259:
```suggestion
  description: A shared memory parallel sparse matrix computations library for the Recursive Sparse Blocks format implementing the Sparse BLAS standard
```

ivan-pi:
#build_preview

awvwgk (approved):

LKedward (approved):

LKedward:
#delete_preview

PR #396, opened 2022-04-15
Add OpenFFT to package index

ivan-pi:
Homepage: http://www.openmx-square.org/openfft/

The code is written in C but includes a Fortran interface. The library is used for DFT electronic structure calculation as part of the OpenMX (http://www.openmx-square.org/) library.


awvwgk (approved):

milancurcic (approved):
Thank you!

PR #398, opened 2022-04-15
Add feed link to HTML head element

mfsch:
Having the news feed linked in the `<head>` element makes it easier to subscribe in some software, as they can automatically detect the address of the feed this way. I’ve used the title “Fortran Newsletter” that’s also used for the feed itself .

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -6,6 +6,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta name="description" content="{{site.description}}">
     <title>{% if page.title %}{{ page.title }} - {% endif %}{{ site.title }}</title>
+    <link type="application/atom+xml" rel="alternate" href="{{ site.baseurl }}/news.xml" title="Fortran Newsletter" />
     <link href="https://fonts.googleapis.com/css?family=Lato:400,700,900" rel="stylesheet">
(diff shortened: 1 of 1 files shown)


milancurcic:
#build_preview

milancurcic:
Thanks @mfsch! Is there an easy tool I could use to check that this works?

mfsch:
I’m using [Miniflux](https://miniflux.app/), but I’ve also just tried out the [Feedbro](https://addons.mozilla.org/en-US/firefox/addon/feedbroreader/) extension for Firefox, which can detect the feed this way, and I think it should also work with something like [this extension](https://chrome.google.com/webstore/detail/rss-feed-reader/pnjaodmkngahhkoihejjehlcdlnohgmp) in Chrome.

milancurcic:
OK, thanks. I installed miniflux but may take some time figuring out how to use it. Can you confirm with the [preview website](https://fortran-lang.org/pr/398/) that the change works as intended? I'm not as familiar with feeds and readers. If yes, I think this PR is good to go.

mfsch:
Miniflux says “Unable to find any subscription.” when trying to add `https://fortran-lang.org/` but it does find the feed when adding `https://fortran-lang.org/pr/398/`.

milancurcic:
Perfect, thanks for confirming, let's merge it.

milancurcic:
#delete_preview

PR #400, opened 2022-04-24
fix dependency of include files under `learn/building_programs` mini-book

p-costa:
I had to do this fix to the dependencies generator, to be able to build a project in the case when the makefile is not in the same directory as the source files. I hope this makes sense!

### DIFF
--- a/learn/building_programs/project_make.md
+++ b/learn/building_programs/project_make.md
@@ -262,7 +262,7 @@ $2 ~ /^["'].+["']$/ {
     # count included files per file to avoid having duplicates in our list
     if (incc[FILENAME,$2]++ == 0) {
         # Add the included file to our list, this might be case-sensitive
-        inc[++ii] = sprintf("$(%s) += %s", FILENAME, $2)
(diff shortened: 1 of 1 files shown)


p-costa:
(I just noticed that I accidentally did not fork the project... I hope this is fine, else I can close and re-submit a proper PR)

awvwgk (approved):
Nice catch.

If you are on it already, could you mend the regex above to also include zeros? `[a-zA-Z][a-zA-Z0-9_]` instead of `[a-zA-Z][a-zA-Z1-9_]`?

awvwgk:
I'm surprised this is actually used, I wrote this guide to show how involved the makefile setup can become and motivate people to use fpm or other high-level build systems with proper Fortran support ;).

p-costa:
> I'm surprised this is actually used, I wrote this guide to show how involved the makefile setup can become and motivate people to use fpm or other high-level build systems with proper Fortran support ;).

*fpm* will come first for me! But I also want to update my naive build workflow using `make` with this nice and robust one.

Issue #232, opened 2021-03-18
More Fortran packages


ivan-pi:
I've found a few more Fortran packages I will try to add one by one:

* [ ] Fortran processing library: http://www.biosfera.dea.ufv.br/fpl/index.html
* [ ] 2DECOMP & FFT: http://www.2decomp.org/decomp_api.html
* [ ] SLICOT: http://slicot.org/
* [ ] FORTRAN Routines for Spectral methods: https://cdm.unimore.it/home/matematica/funaro.daniele/rout.htm
* [ ] NUFFT: https://cims.nyu.edu/cmcl/nufft/nufft.html.bak
* [ ] FFTLog: https://jila.colorado.edu/~ajsh/FFTLog/
* [ ] P3DFFT: https://www.p3dfft.net/
* [ ] Parallel FFT Package: https://cs.sandia.gov/~sjplimp/docs/fft/README.html
* [ ] fftMPI: https://fftmpi.sandia.gov/
* [ ] CSLib: https://github.com/lammps/cslib
* [ ] Quadratures for elastostatic Green's functions on flat triangles: https://cims.nyu.edu/cmcl/quadratures/quadratures.html
* [ ] Quadratures for smooth functions in two and three dimensions: https://cims.nyu.edu/cmcl/quadratures/quadratures.html 
* [ ] FMMLIB2D: https://cims.nyu.edu/cmcl/fmm2dlib/fmm2dlib.html 
* [ ] OpenFFT: http://www.openmx-square.org/openfft/

ivan-pi:
My current plan was only to put them in the package index. 👍

vmagnin:
Maybe some yet unlisted packages could be found in the Curlie directory (the historical DMOZ reborn), especially here:
- https://www.curlie.org/en/Computers/Programming/Languages/Fortran/Libraries
- https://www.curlie.org/en/Computers/Programming/Languages/Fortran/Source_Code

ivan-pi:
Indeed, there are a few interesting ones.

On my PC I found the manual for AxiSEM - a parallel spectal-element solver for 3D wave propagation (http://seis.earth.ox.ac.uk/axisem/).

ivan-pi:
This other seismological code (https://github.com/SeisSol/SeisSol) also contains a fair share of Fortran code.

ivan-pi:
On top of the already long list, I've found pencil-code: https://github.com/pencil-code/pencil-code

certik:
[...]

PR #403, opened 2022-05-06
Add SeisSol to package index

ashirrwad:
Added the SeisSol package listed on Issue: #232  to the package index.

### DIFF
--- a/_data/package_index.yml
+++ b/_data/package_index.yml
@@ -1382,6 +1382,13 @@
   categories: scientific
   license: BSD-3-Clause
 
+- name: SeisSol 
+  github: SeisSol/SeisSol
+  description: A software for the numerical simulation of seismic wave phenomena and earthquake dynamics
+  categories: scientific
+  tags: hpc earthquakes dynamic-rupture
+  license: BSD-3-Clause
+  
 # --- Examples / demos / templates ---
 
 - name: Fortran 2018 examples


milancurcic (approved):
Looks good to me. While at first this looked like a C++ project with Fortran bindings, on closer look it seems to be a mixed C++ and Fortran project.

We've had some discussions in the past whether Fortran bindings to a foreign-language project would be appropriate for the index. While I think they are, I don't think we've reached a consensus about it.

certik (approved):

certik:
I also thought this is a C++ project with legacy Fortran, but it seems it has physics stuff in Fortran and it is still developed in Fortran (see e.g. https://github.com/SeisSol/SeisSol/pull/536), so I would say this package fits into our index.

(If it was just Fortran wrappers, then I would say we should have a category for Fortran wrappers to other packages.)

ivan-pi:
For the record, the developers have told me they would like to remove Fortran soon. See also the comment here: https://github.com/fortran-lang/fortran-lang.org/issues/232#issuecomment-1099727757

ashirrwad:
As the package was still a part of checklist on #232 and the library and its documentation showed references to Fortran usability I figured it was still up for addition. While I did stick to the [guide](https://github.com/fortran-lang/fortran-lang.org/blob/HEAD/PACKAGES.md) the criterias for package addition are a bit ambiguous to me as I am new to the language. If the package doesn't meet the requirements I will add a PR to revert these changes as needed.

PR #408, opened 2022-06-14
Adding QNSTOP and VTMOP ACM-TOMS packages to package-index

thchang:
Hi all,

I wanted to add 2 recent ACM TOMS algorithms from my former PhD advisor's lab to the package index... QNSTOP and VTMOP.  Both published under MIT license.

I'm not why it is showing all these changes, as far as I know, I only added these to the yaml file, but it looks like you have some automated workflow that I do not understand.

Also, I have a question:  perhaps can be replied in a comment?  Prior to 2011, ACM TOMS used it's own license, so all older codes in the vtopt repo are under this license.  Can I still add them?  I'm not sure if the ACM license is open source compatible, I just copied from the ACM website.  It appears to have some extra clauses, a-typical of an open source license.

Example:
  https://github.com/vtopt/vtdirect95/blob/master/LICENSE


milancurcic:
Hi Tyler, thanks for the PR. This is because you opened the PR against the gh-pages branch, which is the build of the site. Make your changes and open a new PR against the master branch instead, then close this PR when ready.

thchang:
Thank-you, makes sense.  I think I meant to open the request from master, but must have clicked the wrong button.

PR #410, opened 2022-06-18
some missing changes and a few new ones from Fortran Discourse discussions

urbanjost:
Some changes that should have been in the last pull request that appear not to be, plus a few additions from recent feedback on
Fortran discourse.

When I look at the file differences they appear as expected, but github was reporting pushing 158 commits; when there were only a few expected.  So far I do not see anything out of place, but not sure what that is indicating yet.


urbanjost:
Something wrong. Withdrawing this. It should only be a few files.

PR #411, opened 2022-06-18
several small updates, most based on comments on Fortran Discourse

urbanjost:
A few errata corrections and some changes discussed on Fortran Discourse.  

For some reason some changes I thought were submitted months ago did not apparently appear.  Those are larger; but resetting and starting from scratch on those because of the time that has passed and just want to start that over again.


certik (approved):

milancurcic (approved):
Thank you!

PR #414, opened 2022-07-24
GSoC’22: Port fortran-lang.org to Sphinx

henilp105:
This pr is to port the Existing site of fortran-lang.org to sphinx.
This GSoC’22 project is done by Henil Shalin Panchal under the Guidance of GSoC mentors Sebastian Ehlert @awvwgk and Rohit Goswami @HaoZeke .
https://summerofcode.withgoogle.com/programs/2022/projects/L28fukIz

Thanks and Regards,
Henil Shalin Panchal


henilp105:
I request all the members of the organisation fortran-lang @fortran-lang @fortran-lang/admins  to please review this pr and add suggestions and comments on this pr. live preview of this pr is available at gh-pages of my repo : https://henilp105.github.io/fortran-lang.org/en/index.html .

Thanks and Regards,
Henil Shalin Panchal

milancurcic:
Thanks, Henil.

On mobile, the Join us section appears before the main section. Here's what the top of the page looks like:

Screenshot_20220724-152455


And here's where the main section begins:

Screenshot_20220724-152526

I will report more issues as I discover them.

henilp105:
Thanks sir, I will surely patch this.

Thanks and Regards,
Henil Shalin Panchal

awvwgk:
The data files like `contributors.json` shouldn't be checked into git, if we are able to generate them in `conf.py`. Instead we should explicitly ignore them in `.gitignore`.

henilp105:
Sir, but they aren't made at the build time, they are generated by workflow (fortran_packages) by using cron .

awvwgk:
Can we generate them in `conf.py` instead? This wouldn't require to continuously push new commits to the repository.

The step can be guarded to be done only if requested, to avoid refetching the data to frequently in local builds, while in CI the environment variable will be set accordingly:

```python
if os.environ.get('SPHINX_GET_DATA', None) == 'True':
    ...
```

henilp105:
We can surely do that sir, but that would lead to increase in the build time by 2.5 mins as this data fetching script takes about 2.5 mins. and in local builds the data is fetched directly from json files from repo by a get request .

awvwgk:
We already have data files checked into the repository, so this should be okay to keep for the time being. Still it would be preferable to not have to check-in those data files in the first place, but we have to store them somewhere, of course.

henilp105:
Sure sir, I will be finding a balanced way out.

Thanks and Regards,
Henil Shalin Panchal

awvwgk:
This has been resolved with the migration to the new repository (see #415)

PR #417, opened 2022-08-07
fixed dependency generation for included files

### LINKED ISSUE #416: Makefile under `An introduction to make` not rebuilding when included files are changed
I hope this is worth opening an issue here.

In addition to *fpm*, I actually use the nice `Makefile` and dependencies generator under [An introduction to make](https://fortran-lang.org/learn/building_programs/project_make) in some projects. The `Makefile` has a dependencies generator which parses Fortran sources to search for dependencies, also for `include`d files, but does not re-build when these include files are changed.

See this example project:
```
$> tree
.
├── gen-deps.awk
├── Makefile
└── src
    ├── bla.f90
    ├── bla-inc.f90


$> cat src/bla.f90
module bla
  implicit none
  private

  public :: say_hello
contains
  subroutine say_hello
    include "bla-inc.f90"
    print *, "Hello, bla!"
  end subroutine say_hello
end module bla


$> cat src/bla-inc.f90
print*,'Hey!'

$> cat Makefile
(...)
# List of all source files
SRCS := src/bla.f90
TEST_SRCS :=
(...)
```
When I `touch src/bla.f90`, the project is correctly rebuilt, but not when I `touch src/bla-inc.f90`. Unfortunately, I'm not proficient enough with GNU make to fix this, but maybe someone can!

p-costa:
My (naive?) attempt to fix #416. Coming from someone who is an'awk'abetic.

Closes #416.

### DIFF
--- a/learn/building_programs/project_make.md
+++ b/learn/building_programs/project_make.md
@@ -259,11 +259,16 @@ $1 ~ /^(#:?)?include$/ &&
 $2 ~ /^["'].+["']$/ {
     # Remove quotes from the included file name
     gsub(/'|"/, "", $2)
-    # count included files per file to avoid having duplicates in our list
-    if (incc[FILENAME,$2]++ == 0) {
-        # Add the included file to our list, this might be case-sensitive
-        inc[++ii] = sprintf("$(%s) += $(%s)", FILENAME, $2)
-    }
+     # Extract path of included file
+     file=FILENAME
+     gsub(".*/","",file)
+     path = FILENAME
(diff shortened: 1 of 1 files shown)


awvwgk:
I think the correct fix is to remove the resolution of the source file, relative includes are not supported by all compilers.

p-costa:
To be done on the new webpage as per Sebastian's suggestion https://github.com/fortran-lang/fortran-lang.org/issues/416#issuecomment-1207750799

PR #421, opened 2023-02-01
fix canonical links for crawlers

henilp105:
This PR is to resolve https://github.com/fortran-lang/webpage/issues/222 , Please refer : https://fortran-lang.discourse.group/t/4942 .

This would add a custom canonical link tag to all the pages of the webpage.

Thanks and Regards,
Henil

CC @awvwgk @certik

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -11,7 +11,7 @@ news:
     url: /news
     feed: /news.xml
 
-url: "https://fortran-lang.github.io"
+url: "https://fortran-lang.org"
 
 plugins:
 


certik:
What will this change do exactly? Will it look like a duplicate?

henilp105:
@certik this patch would show this deprecated site as a duplicate to webpage site , by adding a canonical link which points to the main webpage (shows  a search engine that this is the location of the original content and prioritise this link ).

henilp105:
#build_preview

certik:
Ok, let's try it!

PR #423, opened 2023-04-22
Add redirects to new webpage

### LINKED ISSUE #222: Generate sitemap.xml for crawlers
Can be done via https://sphinx-sitemap.readthedocs.io/en/latest/index.html for sphinx with translations.

henilp105:
This PR is to resolve https://github.com/fortran-lang/webpage/issues/222 , Please refer : https://fortran-lang.discourse.group/t/4942 .

This would add a redirects to new webpage, so it would fix the old url problem in all the search engines.

Thanks and Regards,
Henil

CC @awvwgk @certik


henilp105:
a preview can be seen at : https://henilp105.github.io/fortran-lang.org/

certik (approved):
Let's go ahead with this and see if it fixes the issue. The information is available in this repository, so nothing gets lost.

The absolute highest priority is to get it resolved in the search engines.

milancurcic:
Thanks for this work. Why not just unpublish this site? I assume Bing doesn't index the new one only because this one is up. Redirects are better and more easily handled with a DNS record, but I don't think this is harmful and it's already in. If this resolves the Bing issue, let's please unpublish this site, there's no reason for it to still be served.

certik:
Yes, we should unpublish it. Let's wait a few days and if this does not fix Bing, let's unpublish.

